PR #2, opened 2022-08-10
Feat: KV query tests

foxpy:
Assigned task: [LSC-125](https://p2pvalidator.atlassian.net/browse/LSC-125)

Tests for KV interchain queries:
- balance query;
- delegations query;
- transfers query;
- TX query callback failure;
- KV query callback failure.

Depends on:
- https://github.com/neutron-org/neutron-contracts/pull/18

### DIFF
--- a/src/testcases/interchain_kv_query.test.ts
+++ b/src/testcases/interchain_kv_query.test.ts
@@ -9,6 +9,7 @@ import { wait } from '../helpers/sleep';
 import { AccAddress, ValAddress } from '@cosmos-client/core/cjs/types';
 import { CosmosSDK } from '@cosmos-client/core/cjs/sdk';
 import { InlineResponse20075TxResponse } from '@cosmos-client/core/cjs/openapi/api';
+import { max } from 'lodash';
 
 const getRegisteredQueryResult = (
   cm: CosmosWrapper,
@@ -34,6 +35,19 @@ const getRegisteredQueryResult = (
     };
   }>(contractAddress, {
     get_registered_query: {
+      query_id: queryId,
+    },
+  });
+
+const getKvCallbackStatus = (
+  cm: CosmosWrapper,
+  contractAddress: string,
+  queryId: number,
+) =>
+  cm.queryContract<{
+    last_update_height: number;
+  }>(contractAddress, {
+    kv_callback_stats: {
       query_id: queryId,
     },
   });
@@ -254,7 +268,7 @@ describe('Neutron / Interchain KV Query', () => {
     connectionId = 'connection-0';
     query = {
       1: { updatePeriod: 2, key: '' },
-      2: { updatePeriod: 6, key: '' },
+      2: { updatePeriod: 3, key: '' },
       3: { updatePeriod: 4, key: '' },
       4: { updatePeriod: 3, key: '' },
     };
@@ -562,6 +576,8 @@ describe('Neutron / Interchain KV Query', () => {
         contractAddress,
         transferRecipient,
       );
+      // order is not guaranteed, sorting just for ease of further comparison
+      queryResult.transfers.sort((a, b) => +a.amount - +b.amount);
       expect(queryResult.transfers.length).toEqual(2);
       expect(queryResult.transfers[0]).toEqual({
         recipient: transferRecipient.toString(),
@@ -640,6 +656,90 @@ describe('Neutron / Interchain KV Query', () => {
         denom: cm[2].denom,
         amount: '7000',
       });
+    });
+  });
+
+  describe("Test 'kv' interchain query rollback", () => {
+    test("'kv' callbacks are not being executed on 'tx' queries", async () => {
+      const res = await getKvCallbackStatus(cm[1], contractAddress, 4);
+      expect(res.last_update_height).toEqual(0);
+    });
+
+    test("'kv' callbacks are being executed in 'kv' queries", async () => {
+      const resPrev = await Promise.all(
+        [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+      );
+      // XXX: for some reason, I have to wait for a really long time here
+      await wait(
+        max([1, 2, 3].map((i) => query[i].updatePeriod)) * 3 * BLOCK_TIME,
+      );
+      const res = await Promise.all(
+        [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+      );
+      for (const i in res) {
+        expect(resPrev[i].last_update_height).toBeLessThan(
+          res[i].last_update_height,
+        );
+      }
+    });
+
+    test('enable mock', async () => {
+      await cm[1].executeContract(
+        contractAddress,
+        JSON.stringify({
+          integration_tests_set_kv_query_mock: {},
+        }),
+      );
+    });
+
+    test("'kv' callbacks have stopped updating, contract state is not corrupted", async () => {
+      const start = await Promise.all(
+        [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+      );
+      for (
+        let i = 0;
+        i < max([1, 2, 3].map((i) => query[i].updatePeriod)) * 3;
+        ++i
+      ) {
+        const res = await Promise.all(
+          [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+        );
+        for (const j of res) {
+          expect(j).not.toEqual(0);
+        }
+        await wait(BLOCK_TIME);
+      }
+      const end = await Promise.all(
+        [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+      );
+      expect(start).toEqual(end);
+    });
+
+    test('disable mock', async () => {
+      await cm[1].executeContract(
+        contractAddress,
+        JSON.stringify({
+          integration_tests_unset_kv_query_mock: {},
+        }),
+      );
+    });
+
+    test("now 'kv' callbacks work again", async () => {
+      const resPrev = await Promise.all(
+        [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+      );
+      // XXX: for some reason, I have to wait for a really long time here
+      await wait(
+        max([1, 2, 3].map((i) => query[i].updatePeriod)) * 3 * BLOCK_TIME,
+      );
+      const res = await Promise.all(
+        [1, 2, 3].map((i) => getKvCallbackStatus(cm[1], contractAddress, i)),
+      );
+      for (const i in res) {
+        expect(resPrev[i].last_update_height).toBeLessThan(
+          res[i].last_update_height,
+        );
+      }
     });
   });
 });
--- a/src/testcases/interchain_kv_query.test.ts
+++ b/src/testcases/interchain_kv_query.test.ts
@@ -196,6 +196,7 @@ const registerTransfersQuery = async (
   connectionId: string,
   updatePeriod: number,
   recipient: AccAddress,
+  minHeight: number,
 ) => {
   const res = await cm.executeContract(
     contractAddress,
@@ -205,6 +206,7 @@ const registerTransfersQuery = async (
         connection_id: connectionId,
         update_period: updatePeriod,
         recipient: recipient.toString(),
+        min_height: minHeight.toString(),
       },
     }),
   );
@@ -324,6 +326,7 @@ describe('Neutron / Interchain KV Query', () => {
         connectionId,
         query[4].updatePeriod,
         transferRecipient,
+        1,
       );
     });
   });
@@ -425,9 +428,18 @@ describe('Neutron / Interchain KV Query', () => {
       expect(queryResult.registered_query.query_type).toEqual('tx');
       expect(
         JSON.parse(queryResult.registered_query.transactions_filter),
-      ).toEqual({
-        'transfer.recipient': transferRecipient.toString(),
-      });
+      ).toEqual([
+        {
+          field: 'transfer.recipient',
+          op: 'Eq',
+          value: transferRecipient.toString(),
+        },
+        {
+          field: 'tx.height',
+          op: 'Gte',
+          value: 1,
+        },
+      ]);
       expect(queryResult.registered_query.zone_id).toEqual(
         testState.sdk2.chainID,
       );
--- a/src/testcases/interchain_kv_query.test.ts
+++ b/src/testcases/interchain_kv_query.test.ts
@@ -56,11 +56,8 @@ const waitUntilQueryResultArrives = async (
   cm: CosmosWrapper,
   contractAddress: string,
   queryId: number,
-  numAttempts?: number,
+  numAttempts = 20,
 ) => {
-  if (typeof numAttempts === 'undefined') {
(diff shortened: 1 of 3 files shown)


src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,180 @@
...
+        addr: testState.wallets.demo2.address.toString(), denom: "stake"
+      });
+      expect(queryResult.registered_query.query_type).toEqual("x/bank/GetBalance");
+      expect(queryResult.registered_query.zone_id).toEqual(testState.sdk_2.chainID);
+      expect(queryResult.registered_query.connection_id).toEqual("connection-0");
+      expect(queryResult.registered_query.update_period).toEqual(5);
+      // XXX: can I do this check better than just ensuring height is > 0? Feels like I can.
+      expect(queryResult.registered_query.last_emitted_height).toBeGreaterThan(0);
+      // XXX: `last_submitted_result_local_height` and `last_submitted_result_remote_height`
+      //      are always zero. Is it okay?
```
sotnikov-s, line 129: afaik it's not for the balance query. it would be ok if you used tx query cuz we don't submit the tx query results, but for this one it should have values there. @pr0n00gler am I right?

pr0n00gler: Yes, zero values in this case are not okay

sotnikov-s: ah, wait, we're here in the block where we registered the query, aren't we? if so, zeroes are ok because update period is 5 and the query hasn't run, yet

foxpy: yep, values get updated after query results are submitted, and now I test this behaviour correctly

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+
+const validateQueryRegistration = (
+  res: InlineResponse20075TxResponse,
+  connectionId: string,
+  zoneId: string,
+  updatePeriod: number,
+  queryType: string,
+) => {
+  const attributes = getEventAttributesFromTx({ tx_response: res }, 'wasm', [
+    // TODO: What about checking "kv_keys", too?
```
sotnikov-s, line 130: if it's not too complicated, you can do so. otherwise I think it's out of scope of this test cuz it's mostly the neutron sdk helper's + contract's logic and it should be tested in corresponding unit tests

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+      );
+      expect(res.code).toEqual(0);
+      await wait((query[1].updatePeriod + 1) * BLOCK_TIME);
+
+      await validateBalanceQuery(
+        cm[1],
+        contractAddress,
+        1,
+        testState.sdk2 as CosmosSDK,
+        testState.wallets.demo2.address.toAccAddress(),
```
sotnikov-s, line 478: you've make a sending of 9000 on behalf of `cm[2]`, but then you check whether `testState.wallets.demo2.address.toAccAddress()` equals to query1 result. I don't like two things here:
1. the `testState.wallets.demo2.address.toAccAddress()` doesn't explicitly say that we're checking the cm[2]'s account. I think using `cm2.wallet.address.toString()` would be more clear here;
2. I'd like to see exact values here, like we had 15000 on the balance before, and we've just sent 9000, so we expect it to be 6000. otherwise a mistake can elude you

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+        cm[1],
+        contractAddress,
+        2,
+        testState.sdk2 as CosmosSDK,
+        testState.wallets.val2.address.toAccAddress(),
+      );
+    });
+
+    // TODO: test this query with multiple validators, this is impossible right now
+    //       because we only have one node per network in cosmopark
```
sotnikov-s, line 510: are you planning to do it in this task's scope?

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+        cm[1],
+        contractAddress,
+        3,
+      );
+      expect(interchainQueryResult.delegations[0].amount.amount).toEqual(
+        (3000).toString(),
+      );
+    });
+
+    test('perform icq #4: transfers', async () => {
```
sotnikov-s, line 535: you were supposed to implement kv tests, but this one is a tx test, and the tx queries are covered with tests by me in another file

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+    directQueryResult.data.balances,
+  );
+};
+
+type Query = {
+  updatePeriod: number;
+  key: string;
+};
+
+describe('Neutron / Interchain KV Query', () => {
```
sotnikov-s, line 237: a couple of things that should be tested in additional:
- the contract swallows sudo calls (make changes to the contract's code and add state changing logic to sudo calls, also add a query msg for the contract to read the state. please refer this storage entity usage https://github.com/neutron-org/neutron-contracts/blob/main/contracts/neutron_interchain_queries/src/state.rs#L7);
- with AllowKVCallbacks=false make sure the contract's state hasn't changed;
- getting a query result by a user different from the query owner (optional).

src/testcases/interchain_kv_query.test.ts
foxpy: I agree with point 1, I will fix it.

Point 2 doesn't make much sense for me, since I am testing interchain queries, and I trust ordinary queries to do that. If I cannot trust ordinary queries, then we have bigger problems. Oh, and by the way, your math misses transaction fee, which complicates things a little.

Anyway, I will fix point 1 and, if you ask me again, I will fix point 2 as well.

src/testcases/interchain_kv_query.test.ts
foxpy: Definitely not. I am just leaving this TODO for future generations to care about.

src/testcases/interchain_kv_query.test.ts
foxpy: I agree, but I am not sure if I can do anything about it right now:

https://github.com/neutron-org/neutron-contracts/blob/dd671c5871686cb971a303cfdd0482bef0900065/contracts/neutron_interchain_queries/src/contract.rs#L253-L272

src/testcases/interchain_kv_query.test.ts
sotnikov-s: what I meant and what I wanted to point out is that the validateQuery funcs accept implicit parameters, and if you make a single mistake e.g. pass a wrong address or query there (and there is a lot of addresses and queries and other entities in the code with these index digits etc. which makes it easy to make such a mistake) you will not notice an error. roughly, these funcs actually check whether icq results are same as bank/delegation call results, not whether query result has been updated as you expect

if it still pointless to you, please, don't bother yourself with this and leave as it is. it's just the way I'd like it to be, but it doesn't have to be so

src/testcases/interchain_kv_query.test.ts
sotnikov-s: why not? because of something that @pr0n00gler is supposed to change? is it still relevant?

src/testcases/interchain_kv_query.test.ts
sotnikov-s: could you please elaborate? is it too hard or too premature or what's the reason of not doing it?

src/testcases/interchain_kv_query.test.ts
foxpy: actually I would like to eventually refactor the whole test suite to make it more readable

talking about easy to make mistakes, I could check that `cm[2].wallet`'s balance has decreased during a test, what do you think about it?

src/testcases/interchain_kv_query.test.ts
foxpy: That's way outside of scope of my task, but I will do that if necessary.

src/testcases/interchain_kv_query.test.ts
foxpy: actually I am not even sure. I didn't think I should even touch this code on behalf of my task. or should I actually?

src/testcases/interchain_kv_query.test.ts
foxpy: > the contract swallows sudo calls (make changes to the contract's code and add state changing logic to sudo calls)

I guess I am already doing it?

> AllowKVCallbacks=false

I will have to toggle this setting and restart cosmopark to do that.

> getting a query result by a user different from the query owner

Aren't queries just available to everyone? I am not even using any user account to perform a query.

Sorry for all these questions, I am just not 100% sure what do you mean

src/testcases/interchain_kv_query.test.ts
sotnikov-s: you can take a glance at my code in the neighbour PR, probably it can give you a hint on it

foxpy: has been planning to do that anyway :')

src/testcases/interchain_kv_query.test.ts
sotnikov-s: this is exactly what I did to make my contracts testable by callbacks. remember we had a discussion about where to place the contract — in tests repo or in neutron-contracts one? the discussion took place when I was working on that contract modification

src/testcases/interchain_kv_query.test.ts
sotnikov-s: > Sorry for all these questions, I am just not 100% sure what do you mean

it's totally okay cuz the topic is discussable

> I guess I am already doing it?

you mean the check in the tx query test? you need also to test kv callbacks, and to do so you should modify the contract that helps you in tests (the neutron_interchain_queries one) as I've written above in this thread

src/testcases/interchain_kv_query.test.ts
foxpy: yea it is way out of scope

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+    get_registered_query: {
+      query_id: queryId,
+    },
+  });
+
+const waitUntilQueryResultArrives = async (
+  cm: CosmosWrapper,
+  contractAddress: string,
+  queryId: number,
+  numAttempts?: number,
```
ratik, line 45:
```suggestion
  numAttempts = 20,
```

src/testcases/interchain_kv_query.test.ts
ratik: I think implementing test that require restart of cosmopark with specific parameters looks at least like a separate task

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+    test('get registered icq #1: balance', async () => {
+      await waitUntilQueryResultArrives(cm[1], contractAddress, 1);
+      const queryResult = await getRegisteredQueryResult(
+        cm[1],
+        contractAddress,
+        1,
+      );
+      expect(queryResult.registered_query.id).toEqual(1);
+      expect(queryResult.registered_query.owner).toEqual(contractAddress);
+      // XXX: I could actually check that "key" is correctly derived from contractAddress,
```
ratik, line 341: not a blocker but you can actually, and we have this lib in yarn.lock https://github.com/neutron-org/neutron-integration-tests/blob/master/yarn.lock#L1521

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,633 @@
...
+      // `last_submitted_result_remote_height` to update makes zero sense
+      const queryResult = await getRegisteredQueryResult(
+        cm[1],
+        contractAddress,
+        4,
+      );
+      expect(queryResult.registered_query.id).toEqual(4);
+      expect(queryResult.registered_query.keys.length).toEqual(0);
+      expect(queryResult.registered_query.query_type).toEqual('tx');
+      expect(
```
ratik, line 426: oh this kind of stuff gonna be broken when new filter format is merged. lol

src/testcases/interchain_kv_query.test.ts
ratik: can you just create a task about it so we can decide on planning wether we do it or not

src/testcases/interchain_kv_query.test.ts
ratik: You can for sure modify contract the way it helps you test the scope of the task but keep backward compatibility if possible (so you don't have to fix bash test too)

src/testcases/interchain_kv_query.test.ts
foxpy: Can I do something about it or is it better to just leave things as is?

sotnikov-s: just wait and adapt. so do I

ratik: ok, it's in master

foxpy: fixed!

src/testcases/interchain_kv_query.test.ts
foxpy: okay, I guess now I have implemented this part as well

src/testcases/interchain_kv_query.test.ts
foxpy: [Done](https://p2pvalidator.atlassian.net/browse/LSC-163)

ratik (approved):

oopcode (approved):

oopcode:
tested, everything seems to work fine

src/testcases/interchain_kv_query.test.ts
```diff
@@ -0,0 +1,742 @@
...
+      );
+      expect(queryResult.transfers.length).toEqual(3);
+      expect(queryResult.transfers[2]).toEqual({
+        recipient: transferRecipient.toString(),
+        sender: cm[2].wallet.address.toString(),
+        denom: cm[2].denom,
+        amount: '7000',
+      });
+    });
+  });
```
foxpy, line 657: This test doesn't work anymore. Transfer never arrives again, I guess that's expected behaviour for a stateful relayer?

foxpy: @quasisamurai please leave your opinion here.

context: tx query result arrives, but contract callback handler fails. tx query used to arrive again (in other words, callback was executed again), but it doesn't anymore.

foxpy: This new behaviour is correct, because currently relayer stops submitting tx query results after a failure. The only time when retransmit should happen is when there is network error (still unimplemented).

I should update test accordingly.

quasisamurai: if I understand this correct - this is expected. now we are dropping duplicated txs, even if they're failed

quasisamurai (changes requested):
Please consider update test `Test 'tx' interchain query rollback` with new relayer's behavior: now duplicated (by `queryID` + `txHash`) txs are dropped

quasisamurai:
tested. works nice

quasisamurai (approved):

foxpy:
Will undraft after refactoring

sotnikov-s (approved):

PR #16, opened 2022-09-19
Chore: refactor

foxpy:


src/helpers/cosmos.ts
```diff
@@ -110,15 +109,16 @@ export class CosmosWrapper {
     const txhash = res.data?.tx_response.txhash;
     while (numAttempts > 0) {
       await waitBlocks(this.sdk, 1);
-      try {
-        numAttempts--;
-        const data = await rest.tx.getTx(this.sdk as CosmosSDK, txhash);
+      numAttempts--;
+      const data = await rest.tx
+        .getTx(this.sdk as CosmosSDK, txhash)
+        .catch(() => null);
```
ratik, line 115: mb we can store error and re-throw it at 121?

src/testcases/interchain_kv_query.test.ts
```diff
@@ -465,7 +458,7 @@ describe('Neutron / Interchain KV Query', () => {
       );
       for (
         let i = 0;
-        i < max([2, 3, 4].map((i) => query[i].updatePeriod)) + 1;
+        i < max([2, 3, 4].map((i) => updatePeriods[i])) + 1;
```
ratik, line 461:
```suggestion
        i <= max(Object.values(updatePeriods));
```

src/testcases/interchain_kv_query.test.ts
```diff
@@ -171,18 +171,17 @@ const validateBalanceQuery = async (
...
-// TODO: throw away this type alltogether
-type Query = {
-  updatePeriod: number;
-};
-
 describe('Neutron / Interchain KV Query', () => {
+  const connectionId = 'connection-0';
+  const updatePeriods: { [key: number]: number } = {
+    2: 2,
+    3: 4,
```
NeverHappened, line 178: typo?: `3: 3, 4: 4`

foxpy: It doesn't make any sense actually, it is just a pair of query id and update period

src/helpers/ica.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * getIca queries interchain account address
+ */
+export const getIca = async (
+  cm: CosmosWrapper,
+  contractAddress: string,
+  icaId: string,
+  connectionId: string,
+) => {
+  await waitForIca(cm, contractAddress, icaId, connectionId);
```
NeverHappened, line 39: It seems we always doing double query here: wait for result, and then get same result.
I propose to refactor `waitWithAttempts` func to something like this:
```
// getWhenReady tries to get results until the data is ready
export const getWhenReady = async <T>(
  sdk: any,
  getFunc: () => Promise<T>,
  readyFunc: (t: T) => boolean,
  numAttempts = 20,
): Promise<T> => {
  while (numAttempts > 0) {
    numAttempts--;

    try {
      const data = await getFunc();
      if (readyFunc(data)) {
        return data;
      }
      await waitBlocks(sdk, 1);
    } catch {
      continue;
    }
  }
  throw new Error('getWhenReady: no attempts left');
};
```

I didn't test it!
But this allows to separate check if ready from getting and to return result simultaneously.

foxpy: That means I will have to redo `waitWithAttempts()` somehow to make its predicate not only return completion status, but also return queried data.

I don't have any ideas on how to do that without writing too much code, and tests wouldn't really speed up after saving 1 ms or less, so it doesn't seem to be worthy for me to refactor these functions.

They are easy to read and they work well enough.

Unless you have an idea on how to do that cleanly?

foxpy: oops, sorry, I didn't read your snippet thoroughly :facepalm: 

I will investigate if it is possible to refactor the waiting functions

src/testcases/interchaintx.test.ts
```diff
@@ -406,20 +281,20 @@ describe('Neutron / Interchain TXs', () => {
...
+      await getWithAttempts(
         cm1.sdk,
-        async () =>
-          (
-            await cm1.listIBCChannels()
-          ).channels.filter((c) => c.channel_id == 'channel-3')[0].state ==
-          'STATE_OPEN',
+        () => cm1.listIBCChannels(),
+        (channels) =>
+          channels.channels.filter((c) => c.channel_id == 'channel-3')[0]
```
NeverHappened, line 296: There is `array.find` function that you can use

src/testcases/interchain_tx_query.test.ts
```diff
@@ -297,11 +302,17 @@ describe('Neutron / Interchain TX Query', () => {
       const res = await cm2.msgSend(watchedAddr3, amountToAddr3_2.toString());
       expectedIncomingTransfers++;
       expect(res.code).toEqual(0);
+      // initiate query before relayer has any chance to submit query data
+      const depositsPromise = queryRecipientTxs(
```
NeverHappened, line 306: I don't understand why can't we move this to line 315 where we wait for it?

foxpy: it literally says `// initiate query before relayer has any chance to submit query data`

README.md
```diff
@@ -36,7 +38,8 @@ yarn test:interchain_tx_query # interchain tx query test
 yarn test:interchain_kv_query # interchain tx query test
 ```
 
-NOTE: To speed up tests you can run the cosmopark by youself with `make start-cosmopark` in the setup dir. To run test with the already running cosmopark use `NO_DOCKER` env variable.
+NOTE: To speed up tests you can run the cosmopark by youself with `make start-cosmopark` in the setup dir. To run test
+with the already running cosmopark use `NO_DOCKER` env variable.
```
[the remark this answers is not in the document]
foxpy: oops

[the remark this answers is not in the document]
foxpy: actually not oops, seems like it has been automatically capped at 120 characters per line.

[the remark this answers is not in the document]
sotnikov-s: Wait, we have even longer lines in other repos. It means the linter (if it's the linter who's done it) doesn't know its place and tries to control `.md` files whereas it's meant to take care about `.ts` files only. The interesting fact is that the lines are the same they are in main, but only in your PR they've been changed, means it's most likely your linter settings that's made it. If it's not the linter, then IDE maybe, but anyway it seems redundant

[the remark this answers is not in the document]
foxpy: That was my IDE. I have canceled these changes.

oopcode (approved):

sotnikov-s (approved):

PR #21, opened 2022-10-12
feat: introduce packet id to check acknowledgement for concrete packet id

NeverHappened:
related to [this task](https://p2pvalidator.atlassian.net/browse/NTRN-221)
- use sequence id to query for needed acknowledgement result instead of just the last one


oopcode (approved):

src/testcases/interchaintx.test.ts
```diff
@@ -375,6 +373,7 @@ const waitForAck = (
   cm: CosmosWrapper,
   contractAddress: string,
   icaId: string,
+  sequenceId: number,
```
swelf19, line 376: Why do not return the ack in the `waitForAck` function, instead of making separate call `getAck`

swelf19 (dismissed):

pr0n00gler (approved):

PR #24, opened 2022-10-14
fix: updated test according new contracts method name

swelf19:
contracts pr: https://github.com/neutron-org/neutron-contracts/pull/43


foxpy:
@swelf19 CI basically asks you to run `yarn fmt`

foxpy (changes requested):

foxpy:
Why do you target main instead of audit fixes branch?

PR #39, opened 2022-11-18
printing versions of services after starting containers

oldremez:
1. After starting containers, test code request versions of the binaries inside the containers so we can be sure we're running the version we expect to run
2. Makefile fixed to pass the flags to neutron-query-relayer builder container

### DIFF
--- a/src/helpers/env.ts
+++ b/src/helpers/env.ts
@@ -23,6 +23,7 @@ export const setup = async (host: string) => {
   } catch (e) {}
   console.log('Starting container... it may take long');
   execSync(`cd setup && make start-cosmopark`);
+  showVersions();
   await waitForHTTP(host);
   await waitForChannel(host);
   alreadySetUp = true;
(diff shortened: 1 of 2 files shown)


src/helpers/env.ts
```diff
@@ -69,3 +70,31 @@ export const waitForChannel = async (
   }
   throw new Error('No channel opened');
 };
+
+export const showVersions = () => {
```
albertandrejev, line 74: It would be great to show `commit` for `neutrond` and `gaiad` as well.

albertandrejev (dismissed):

src/helpers/env.ts
```diff
@@ -79,13 +79,16 @@ export const showVersions = () => {
   const servicesAndGetVersionCommands = [
     [
       'neutrond',
-      'cd setup && docker compose exec neutron-node /go/bin/neutrond version',
+      `cd setup && docker compose exec neutron-node sh -c "/go/bin/neutrond version version --long | grep '^commit\\|^version'"`,
```
albertandrejev, line 82: you added `version` twice by mistake.

oldremez: Actually, I decided to try to move the complexity from weird-ass shell one-liner to ts code but how it even worked lol

albertandrejev (approved):

ratik (approved):

PR #42, opened 2022-11-24
feat: NTRN-267. ICQ removal after timeout

albertandrejev:
**Summary**

It is possible by anyone to delete ICQ after submission timeout and receive deposit from it. We need tests for this behavior.

**DoD**

Following test cases should be implemented:

- attempt to delete ICQ before submission timeout should fail;
- attempt to delete ICQ after submission timeout should succeed (and funds should be returned);
- change of timeout via governance should be reflected (in case if it is impossible to use governance in current integration tests, TODO comment should be left);
- change of deposit should be reflected, it is especially important to check that the whole deposit is always being returned (in case if it is impossible to use governance in current integration tests, TODO comment should be left).

**Related PRs**
https://github.com/neutron-org/neutron/pull/95
https://github.com/neutron-org/neutron-dev-contracts/pull/1


oopcode (approved):

.gitignore
```diff
@@ -13,4 +12,4 @@ contracts/artifacts/checksums.txt
 .vscode
 
 write-file-atomic*
-*.wasm
```
ratik, line 16: are we going to store binaries again?

albertandrejev: only for DAO contracts

foxpy: please don't :crying_cat_face:

albertandrejev: Yes, it should move to the meta-repo. Let's see

src/testcases/interchain_kv_query.test.ts
```diff
@@ -116,22 +117,100 @@ const registerBalanceQuery = async (
...
+  await cm.executeProposal(proposalId, wallet.address.toString());
+
+  await waitBlocks(cm.sdk, 5);
+
+  /*await getWithAttempts(
+    cm.sdk,
+    () => cm.queryProposal(proposalId.toString()),
+    async (response) => response.proposal?.status === 'PROPOSAL_STATUS_PASSED',
+    20,
+  );*/
```
foxpy, line 189: Wait, why is it commented out? I see you wanted to wait for actual proposal approval, but instead you just wait for N blocks to pass. Please remove fixed sleeps and wait until proposal is accepted.

foxpy (changes requested):

src/testcases/interchain_kv_query.test.ts
albertandrejev: fixed

foxpy (commented):
As much as I would like to approve it, two last tests in KV query test suite are failing.

albertandrejev:
> As much as I would like to approve it, two last tests in KV query test suite are failing.

Yes, I know. I said about it in my message. I believe these tests should be fixed in the https://github.com/neutron-org/neutron-integration-tests/pull/51
This happens because of incompatibility in the old DAO contracts the last changes in neutron.

foxpy (approved):
Let's get it merged for now

quasisamurai (approved):
tested ack

PR #46, opened 2022-11-29
fix: NTRN-278. Fix integration tests to work with interchain security

albertandrejev:
Make integration tests work with latest version of neutron with CCV modules.
It should work with `main` version of `neutron`.


oopcode (approved):
utACK

PR #47, opened 2022-12-01
Show contracts hashes

oldremez:


ratik (approved):

src/helpers/cosmos.ts
```diff
@@ -442,3 +447,20 @@ export const getSequenceId = (rawLog: string | undefined): number => {
...
+    hashes[NeutronContract[key]] = crypto
+      .createHash('sha256')
+      .update(binary)
+      .digest('hex');
+  }
+  return hashes;
+};
+
+const getContractBinary = async (fileName: string): Promise<Buffer> => {
+  const contractPath = process.env.CONTRACTS_PATH || './contracts/artifacts';
```
albertandrejev, line 464: I think we can move this line on line 25, or somewhere there.

src/helpers/env.ts
```diff
@@ -113,3 +116,14 @@ export const showVersions = () => {
...
+const showContractsHashes = async () => {
+  const hashes = await getContractsHashes();
+
+  let result = 'Contracts hashes:\n';
+  for (const key of Object.keys(hashes)) {
+    result = result.concat(`${hashes[key]} ${key}\n`);
+  }
+
+  console.log(result);
+};
```
albertandrejev, line 129: I think this code is too complicated. What do you think to move content of the `getContractsHashes` here and simply print output? Something like:
```typescript
const showContractsHashes = async (): Promise<void> => {
  for (const key of Object.keys(NeutronContract)) {
    const binary = await getContractBinary(NeutronContract[key]);
    const hash = crypto
      .createHash('sha256')
      .update(binary)
      .digest('hex');
     console.log(`${hash} ${key}\n`);
  }
};
```

src/helpers/cosmos.ts
oldremez: Yup, reasonable, will do

src/helpers/env.ts
oldremez: I wrote this while keeping in mind that we will (1) implement comparing hashes with what we have on github and (2) adapt the code to use different environments (k8s and so on) further, so that show/get hashes functions will be separated anyway eventually.

src/helpers/cosmos.ts
```diff
@@ -17,11 +17,13 @@ import { CosmosSDK } from '@cosmos-client/core/cjs/sdk';
 import ICoin = cosmos.base.v1beta1.ICoin;
 import { ibc } from '@cosmos-client/ibc/cjs/proto';
 import IHeight = ibc.core.client.v1.IHeight;
+import crypto from 'crypto';
 
 export const NEUTRON_DENOM = process.env.NEUTRON_DENOM || 'stake';
 export const COSMOS_DENOM = process.env.COSMOS_DENOM || 'uatom';
 export const IBC_RELAYER_NEUTRON_ADDRESS =
   'neutron1mjk79fjjgpplak5wq838w0yd982gzkyf8fxu8u';
+const contractPath = process.env.CONTRACTS_PATH || './contracts/artifacts';
```
albertandrejev, line 26: Can you use same case convention for constant names please?

oldremez:
Oh btw fixed readme here (removed branch specification since we changed the main branch for contracts to `main` that's the default one in script) just not to create a separate PR for that small change

src/helpers/cosmos.ts
oldremez: 🤦‍♂️

sotnikov-s (approved):

albertandrejev (approved):

ratik (approved):

PR #48, opened 2022-12-02
fix: NTRN-267. Interchaintx tests. Add counterparty chain connection to be established

albertandrejev:
Because of delay in the connection creation using query relayer it is required to change approach used to detect connection established event.

### DIFF
--- a/src/helpers/wait.ts
+++ b/src/helpers/wait.ts
@@ -1,8 +1,8 @@
 import { rest } from '@cosmos-client/core';
 
-export const wait = async (n: number) =>
+export const wait = async (seconds: number) =>
   new Promise((r) => {
-    setTimeout(() => r(true), n);
+    setTimeout(() => r(true), 1000 * seconds);
   });
 
 /*
@@ -19,7 +19,7 @@ export const getRemoteHeight = async (sdk: any) => {
 export const waitBlocks = async (sdk: any, n: number) => {
   const targetHeight = (await getRemoteHeight(sdk)) + n;
   for (;;) {
-    await wait(10);
+    await wait(1);
     const currentHeight = await getRemoteHeight(sdk);
     if (currentHeight >= targetHeight) {
       break;
--- a/src/helpers/env.ts
+++ b/src/helpers/env.ts
@@ -61,7 +61,12 @@ export const waitForChannel = async (
       const r = await axios.get(`${host}/ibc/core/channel/v1/channels`, {
         timeout: 1000,
       });
-      if (r.data.channels.length > 0) {
+      if (
+        r.data.channels.length > 0 &&
+        r.data.channels.every(
+          (channel: any) => channel.counterparty.channel_id !== '',
+        )
+      ) {
         return;
       }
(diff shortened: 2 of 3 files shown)


swelf19 (approved):
tested
main - fails
fix/NTRN-267-fix-interchaintx-tests - works

swelf19 (approved):

sotnikov-s (approved):
lgtm, integration tests pass

src/helpers/wait.ts
```diff
@@ -19,7 +19,7 @@ export const getRemoteHeight = async (sdk: any) => {
 export const waitBlocks = async (sdk: any, n: number) => {
   const targetHeight = (await getRemoteHeight(sdk)) + n;
   for (;;) {
-    await wait(10);
+    await wait(1);
```
foxpy, line 22: \*cries in endpoint access logs\*

src/helpers/wait.ts
```diff
@@ -1,8 +1,8 @@
 import { rest } from '@cosmos-client/core';
 
-export const wait = async (n: number) =>
+export const wait = async (seconds: number) =>
   new Promise((r) => {
-    setTimeout(() => r(true), n);
+    setTimeout(() => r(true), 1000 * seconds);
   });
```
foxpy, line 6: I speculate this will make tests slightly slower since now `waitBlocks()` will have to wait one more second each time it doesn't spot new block in 1sec window.

Should we revert `wait` and `getRemoteHeight` back or test my assumptions instead? I can run tests if you are interested.

oldremez: I might be wrong in my guess tho

foxpy: the problem is, I do not propose an optimization, I propose to revert code which adds no value, only makes things potentially slower :thinking:

foxpy: By "adds no value" I mean two things:
- it doesn't introduce any features and doesn't change existing features;
- it doesn't make tests run faster, in fact, it makes them potentially slower.

On the other hand, executing up to 100 HTTP requests per second might slow down small machines, especially single vCPU VMs (do we plan to use these to run tests on?).

foxpy: Soooo, if we are changing this anyway, why don't make `wait` accept milliseconds and `waitBlocks` sleep for 100 milliseconds? Sounds like a good compromise for me.

foxpy: It is pointless to argue on speculated data, so I decided to run benchmarks.
On my machine `yarn test` runs for 9m46.931s with 1 sec delays and 7m54.878s with 100 msec delays.

We add almost two minutes of a delay without adding any new tests =(

Please revert these changes, we have plans to completely overhaul tests again in future anyway, making them much slower right now just for sake of more beautiful code is not something I would like to accept.

oopcode: sorry to interrupt; we'll probably refactor the tests a lot soon & try to speed them up, so i'm not sure it's worth arguing in this particular case. i'll merge this pr

PR #53, opened 2022-12-23
feat: Implement burn-based vesting for treasury. NTRN-308

albertandrejev:
Treasury tokens are vested based on on-chain activity: burnt_tokens * a_multiplier. The multiplier is an linear function of the supply: while initially, one burnt tokens equals multiple NTRN tokens made liquid, the flow of new tokens into the Treasury progressively slows down until the tokens supply is exhausted and the tokenomy becomes deflationary.

Depends on:
* https://github.com/neutron-org/neutron-dao/pull/15
* https://github.com/neutron-org/neutron/pull/117

### DIFF
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -855,6 +855,20 @@ export class CosmosWrapper {
     return req.data;
   }
 
+  async queryTotalBurnedNeutronsAmount(): Promise<TotalBurnedNeutronsAmountResponse> {
+    try {
+      const req = await axios.get<TotalBurnedNeutronsAmountResponse>(
+        `${this.sdk.url}/neutron/feeburner/total_burned_neutrons_amount`,
+      );
+      return req.data;
+    } catch (e) {
+      if (e.response?.data?.message !== undefined) {
+        throw new Error(e.response?.data?.message);
+      }
+      throw e;
+    }
+  }
+
   async queryTotalSupplyByDenom(
     denom: string,
   ): Promise<TotalSupplyByDenomResponse> {
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -876,20 +876,6 @@ export class CosmosWrapper {
     return req.data;
   }
 
-  async queryTotalBurnedNeutronsAmount(): Promise<TotalBurnedNeutronsAmountResponse> {
-    try {
-      const req = await axios.get<TotalBurnedNeutronsAmountResponse>(
-        `${this.sdk.url}/neutron/feeburner/total_burned_neutrons_amount`,
-      );
-      return req.data;
-    } catch (e) {
-      if (e.response?.data?.message !== undefined) {
-        throw new Error(e.response?.data?.message);
-      }
-      throw e;
-    }
-  }
-
   async queryTotalSupplyByDenom(
     denom: string,
   ): Promise<TotalSupplyByDenomResponse> {
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -126,6 +126,13 @@ export type AckFailuresResponse = {
   };
 };
 
+// TotalBurnedNeutronsAmount is the response model for the feeburner's  total-burned-neutrons.
+export type TotalBurnedNeutronsAmount = {
+  total_burned_neutrons_amount: {
+    coins: ICoin[];
+  };
+};
+
 // Failure represents a single contractmanager failure
 type Failure = {
   address: string;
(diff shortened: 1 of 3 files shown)


src/testcases/treasury.test.ts
```diff
@@ -41,45 +48,92 @@ describe('Neutron / Treasury', () => {
...
+
+        burnedCoins = await getBurnedCoinsAmount(cm, cm.denom);
+        await cm.executeContract(
+          treasury,
+          JSON.stringify({
+            distribute: {},
+          }),
+        );
+        stats = await cm.queryContract(treasury, { stats: {} });
+        expect(stats.total_processed_burned_coins).toEqual(burnedCoins);
```
sotnikov-s, line 99: we test here whether stats' processed burned coins can properly follow the burned coins amount taken from neutron, right? and we assume that with each `executeContract(distribute)` new coins get burned, so this is why the test is named `burned coins increment`. if I get it right, it would be good to also test whether `burnedCoins` increases between calls

src/testcases/treasury.test.ts
```diff
@@ -97,13 +151,13 @@ describe('Neutron / Treasury', () => {
...
       });
       test('payout by unauthorized', async () => {
-        treasury = await setupTreasury(cm, {
-          owner: testState.wallets.neutron.demo1.address.toString(),
-          distributionRate: '0.23',
-          minPeriod: 1000,
-          distributionContract: dsc,
-          reserveContract: reserve,
-        });
+        // treasury = await setupTreasury(cm, {
```
sotnikov-s, line 154: can be removed

src/testcases/treasury.test.ts
```diff
@@ -116,18 +170,59 @@ describe('Neutron / Treasury', () => {
           ),
         ).rejects.toThrow(/unauthorized/);
       });
+
+      test('burned coins amount u32 overflow', async () => {
+        // treasury = await setupTreasury(cm, {
```
sotnikov-s, line 175: can be removed

src/testcases/treasury.test.ts
```diff
@@ -116,18 +170,59 @@ describe('Neutron / Treasury', () => {
...
+        //   owner: testState.wallets.neutron.demo1.address.toString(),
+        //   distributionRate: '0.0',
+        //   minPeriod: 1,
+        //   distributionContract: dsc,
+        //   reserveContract: reserve,
+        // });
+        await cm.msgSend(treasury, '100000');
+        // u32::MAX
+        await cm.simulateFeeBurning(4_294_967_295);
```
sotnikov-s, line 184: but it's not an overflow

src/testcases/treasury.test.ts
```diff
@@ -145,35 +240,52 @@ describe('Neutron / Treasury', () => {
...
-        expect(stats).toEqual({
-          total_received: '10000000',
-          total_distributed: '2100000',
-          total_reserved: '7900000',
-        });
+        expect(stats).toEqual(
+          expect.objectContaining({
+            total_distributed: '42013',
+            total_reserved: `${
+              158050 + parseInt(treasuryStats.total_reserved)
```
sotnikov-s, line 263: could you please describe these magic numbers somehow? if things go the way the test fails here, it will be complicated to understand why we expect here exactly these values

sotnikov-s (changes requested):

src/helpers/cosmos.ts
```diff
@@ -548,6 +573,20 @@ export class CosmosWrapper {
     }
   }
 
+  async queryTotalBurnedNeutronsAmount(): Promise<TotalBurnedNeutronsAmount> {
+    try {
+      const req = await axios.get<TotalBurnedNeutronsAmount>(
+        `${this.sdk.url}/neutron/feeburner/totalburnedneutronsamount`,
```
foxpy, line 579: the path has changed:
```suggestion
        `${this.sdk.url}/neutron/feeburner/total_burned_neutrons_amount`,
```

src/testcases/treasury.test.ts
albertandrejev: agree, wrong test name, it checks `safe_burned_coins_for_period` function in the contract.

src/testcases/treasury.test.ts
albertandrejev: No, because of u32 limitations we can not process more then u32::MAX burned coins at a time. So we we have burned more then u32::MAX coins then we need to call contract at least two times to process all burned coins. See https://github.com/neutron-org/neutron-dao/blob/306c06e1b39e90b6a1d41ee114a32b6b751b7a37/contracts/tokenomics/treasury/src/vesting.rs#L35

albertandrejev: wrong answer :point_up: , mixed up it with `burned coins amount u32 safe calculation` test.

albertandrejev: It was implemented, see line 91

PR #64, opened 2023-01-15
feat: add test for subDAO config update via proposal

sotnikov-s:
this PR prettifies `subdao.test.ts` and adds to it an additional test that updates subDAO config using a proposal

**contracts:** https://github.com/neutron-org/neutron-dao/pull/29
**tests result:** https://github.com/neutron-org/neutron-tests/actions/runs/3925298149.

### DIFF
--- a/src/testcases/subdao.test.ts
+++ b/src/testcases/subdao.test.ts
@@ -6,10 +6,12 @@ import {
   createBankMassage,
   NeutronContract,
 } from '../helpers/cosmos';
+import { InlineResponse20075TxResponse } from '@cosmos-client/core/cjs/openapi/api';
 import {
   TimelockConfig,
   TimelockProposalListResponse,
   TimeLockSingleChoiceProposal,
+  SubDaoConfig,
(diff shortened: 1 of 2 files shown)


src/testcases/subdao.test.ts
```diff
@@ -853,9 +662,57 @@ describe('Neutron / Subdao', () => {
     });
   });
 
+  describe('Subdao: Proposals and access', () => {
+    let proposal_id: number;
+    test('Update config: Unauthorized', async () => {
```
swelf19, line 667: i would suggest to test that `not timelocked` proposal is  forbidden as well, but to test it we need to instantiate a second propsal module with policy `AnyoneMayPropose`. We can make it as a separate task

swelf19 (dismissed):

pr0n00gler (commented):
Fix linter errors, pls

pr0n00gler (approved):

swelf19 (approved):

PR #66, opened 2023-01-16
feat: snake-case subdao #NTRN-322

ratik:
[Task](https://p2pvalidator.atlassian.net/browse/NTRN-322)
Related [PR](https://github.com/neutron-org/neutron-dao/pull/26)
and [Run](https://github.com/neutron-org/neutron-tests/actions/runs/3929068160)


NeverHappened (approved):

sotnikov-s (approved):

swelf19:
I think we need to make snake_case format not only for messages, but for responses and queires as well, i'm not sure the all necessary types are updated. `grep -ri 'pub enum' -B4 *` you can see a lot of types with no snake_case attribute

PR #68, opened 2023-01-18
fix: race during TimeoutSudo in simple test

swelf19:
What is going on here
We disconnect hermes from the docker network. Wait some time. Gaiad blocks are being built, but neutron knows nothing  about it since hermes is not reachable. We execute tx on the neutron side and connect hermes back to the network. The timeout height has passed, hermes have to send timeoutAck and trigger the timeoutSudoHandler.

https://github.com/neutron-org/neutron-tests/actions/runs/4292498762


src/testcases/common_localcosmosnet.ts
```diff
@@ -12,6 +12,20 @@ export const disconnectValidator = async (name: string) => {
   return stdout;
 };
 
+export const disconnectHermes = async () => {
+  const { stdout } = await exec(
+    `docker network setup_default disconnect setup-hermes-1`,
```
NeverHappened, line 17: So we talked earlier about disconectHermes and connectHermes function.
So i tried to run it locally with debug and got code 1 from this command
```  const res = await exec(
    `docker network setup_default disconnect setup-hermes-1`,
  );
  const { stdout } = res;
  stdout.on('data', (data) => {
    console.log(`Received chunk ${data}`);
  });
  res.on('close', (data) => {
    console.log(`close code: ${data}`);
  });
  console.log('disconnect');
  // console.log(stdout);
  return stdout;```

Tests still run okay locally tho)

NeverHappened (changes requested):

NeverHappened:
also pr in draft, but task in review

NeverHappened (approved):

PR #75, opened 2023-01-31
feat: NTRN-332. Test token factory

albertandrejev:
Osmosis has the token factory module that allows anyone to create new native tokens. We want to have it in Neutron, so we need to see if it’s possible to easily include it to our codebase

Depends on:
https://github.com/neutron-org/neutron/pull/131


foxpy (approved):

albertandrejev:
> it would be cool to see some actual tests tbh. like, at least testing the basic methods of the modul

albertandrejev:
> it would be cool to see some actual tests tbh. like, at least testing the basic methods of the module?

I will add

src/testcases/parallel/tokenfactory.test.ts
```diff
@@ -0,0 +1,325 @@
...
+}
+
+interface AuthorityMetadata {
+  readonly authority_metadata: { readonly Admin: string };
+}
+
+describe('Neutron / Tokenfactory', () => {
+  let testState: TestStateLocalCosmosTestNet;
+  let cmNeutron: CosmosWrapper;
+  let owner_wallet: Wallet;
```
foxpy, line 45:
```suggestion
  let ownerWallet: Wallet;
```

foxpy (dismissed):

pr0n00gler (approved):

pr0n00gler (approved):

ratik (approved):

PR #79, opened 2023-02-02
#NTRN-347 Removing artifacts from repo

oldremez:
1. Artifacts are removed and `*.wasm` is added to `.gitignore`
2. Added script to download artifacts from Google Cloud bucket
3. Got rid of the artifacts duplication
4. Minor init shell scripts improvements

Before testing don't forget to update yarn dependencies

How to test:
1. update yarn dependencies according to readme
2. download artifacts according to readme
3. run tests as usual, make sure they passed

You can also check different scenarios of using the script besides default one (help, verbose mode, different branch/directory)


foxpy (approved):

download_artifacts.js
```diff
@@ -0,0 +1,165 @@
...
+  verboseLog(`Getting latest commit by url:\n${url}`);
+  const resp = (await axios.get(url)).data;
+  return resp['commit']['sha'];
+};
+
+// -------------------- STORAGE --------------------
+
+const getChecksumsTxt = async (
+  repo_name,
+  storage_addr,
```
swelf19, line 50: Is not used anywhere

download_artifacts.js
```diff
@@ -0,0 +1,165 @@
...
+      'contracts repos to download artifacts for',
+    ),
+  );
+
+  program.parse();
+
+  const options = program.opts();
+  const branch_name = options.branch || DEFAULT_BRANCH;
+  const dest_dir = options.dir || DEFAULT_DIR;
+  const repos_to_download = program.args;
```
swelf19, line 154: may be use a default set of repos or return an error in case of empty list? Nothing happens when a run `nodejs download_artifacts.js` and it's a bit confusing

swelf19 (approved):

download_artifacts.js
oldremez: Removed it

download_artifacts.js
oldremez: Just made the `repo` argument required

download_artifacts.js
```diff
@@ -0,0 +1,416 @@
...
+    ? error
+    : new Error(
+        'getWithAttempts: no attempts left. Latest get response: ' +
+          (data === Object(data) ? JSON.stringify(data) : data).toString(),
+      );
+};
+
+const askForRewrite = async (file_name) => {
+  const ok = await yesno({
+    question: `File ${file_name} already exists, do you want to rewrite? (if yes, all files are to be rewritten)`,
```
foxpy, line 78:
```suggestion
    question: `File ${file_name} already exists, do you want to overwrite it? (if yes, all further differing files will be overwritten)`,
```

I believe this version would be more concise.

oldremez: Agree

PR #81, opened 2023-02-10
For parallel tests

MrTalecky:
Task [NTRN-312](https://p2pvalidator.atlassian.net/jira/software/projects/NTRN/boards/121?selectedIssue=NTRN-312)

Tests work when run with the `NO_DOCKER=1 `flag

A new command has been added to package.json to run two test suites in parallel:
`test:parallel `

These two test suites (`tokenomics` and `simple`) use the generated wallets, and run in parallel. 
You can also run `treausury` together with it - it still uses the old wallets, but it can run in parallel with `tokenomics` and `simple`.

[NTRN-312]: https://p2pvalidator.atlassian.net/browse/NTRN-312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -235,7 +235,7 @@ export class CosmosWrapper {
               mode: proto.cosmos.tx.signing.v1beta1.SignMode.SIGN_MODE_DIRECT,
             },
           },
-          sequence: sequence,
+          sequence,
         },
       ],
       fee,
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -217,6 +217,7 @@ export class CosmosWrapper {
     msgs: T[],
     numAttempts = 10,
     mode: rest.tx.BroadcastTxMode = rest.tx.BroadcastTxMode.Async,
+    sequence: number = this.wallet.account.sequence,
   ): Promise<CosmosTxV1beta1GetTxResponse> {
     const protoMsgs: Array<google.protobuf.IAny> = [];
     msgs.forEach((msg) => {
@@ -234,7 +235,7 @@ export class CosmosWrapper {
               mode: proto.cosmos.tx.signing.v1beta1.SignMode.SIGN_MODE_DIRECT,
             },
           },
-          sequence: this.wallet.account.sequence,
+          sequence: sequence,
         },
       ],
       fee,
@@ -273,6 +274,7 @@ export class CosmosWrapper {
       }
     }
     error = error ?? new Error('failed to submit tx');
+    console.log(error);
     throw error;
   }
 
@@ -406,13 +408,22 @@ export class CosmosWrapper {
       gas_limit: Long.fromString('200000'),
       amount: [{ denom: this.denom, amount: '1000' }],
     },
+    sequence: number = this.wallet.account.sequence,
   ): Promise<InlineResponse20075TxResponse> {
     const msgSend = new proto.cosmos.bank.v1beta1.MsgSend({
       from_address: this.wallet.address.toString(),
       to_address: to,
       amount: [{ denom: this.denom, amount }],
     });
-    const res = await this.execTx(fee, [msgSend]);
+    console.log('sequence in msgSend from test');
+    console.log(sequence);
+    const res = await this.execTx(
+      fee,
+      [msgSend],
+      10,
+      rest.tx.BroadcastTxMode.Block,
+      sequence,
+    );
     return res?.tx_response;
   }
 
@@ -1203,7 +1214,16 @@ export const mnemonicToWallet = async (
   }
   return new Wallet(address, account, pubKey, privKey, addrPrefix);
 };
-
+export const createAddress = async (mnemonicQA?: string) => {
+  const privKey = new proto.cosmos.crypto.secp256k1.PrivKey({
+    key: await cosmosclient.generatePrivKeyFromMnemonic(mnemonicQA),
+  });
+  console.log('this is mnemonic from neutron');
+  console.log(mnemonicQA);
+  const pubKey = privKey.pubKey();
+  const address = cosmosclient.AccAddress.fromPublicKey(pubKey).toString();
+  return address;
+};
 export const getSequenceId = (rawLog: string | undefined): number => {
   if (!rawLog) {
     throw 'getSequenceId: empty rawLog';
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -708,8 +708,27 @@ export class CosmosWrapper {
       sender,
     );
   }
-  async getSeq() {
-    return Number(this.wallet.account.sequence);
+  async getSeq(
+    sdk: cosmosclient.CosmosSDK,
+    address: cosmosclient.AccAddress,
+  ): Promise<number> {
+    const account = await rest.auth
+      .account(sdk, address)
+      .then((res) =>
+        cosmosclient.codec.protoJSONToInstance(
+          cosmosclient.codec.castProtoJSONOfProtoAny(res.data.account),
+        ),
+      )
+      .catch((e) => {
+        console.log(e);
+        throw e;
+      });
+
+    if (!(account instanceof proto.cosmos.auth.v1beta1.BaseAccount)) {
+      throw new Error("can't get account");
+    }
+
+    return account.sequence;
   }
   /**
    * submitSoftwareUpgradeProposal creates proposal.
@@ -1219,7 +1238,7 @@ export const createAddress = async (mnemonicQA?: string) => {
     key: await cosmosclient.generatePrivKeyFromMnemonic(mnemonicQA),
   });
   const pubKey = privKey.pubKey();
-  const address = cosmosclient.AccAddress.fromPublicKey(pubKey).toString();
+  const address = cosmosclient.AccAddress.fromPublicKey(pubKey);
   return address;
 };
 export const getSequenceId = (rawLog: string | undefined): number => {
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -192,6 +192,9 @@ cosmosclient.codec.register(
 );
 
 export class CosmosWrapper {
+  sendTokensWithRetry(cm: CosmosWrapper, address: string, arg2: string, sequence: number) {
+    throw new Error('Method not implemented.');
+  }
   sdk: cosmosclient.CosmosSDK;
   blockWaiter: BlockWaiter;
   wallet: Wallet;
@@ -708,7 +711,9 @@ export class CosmosWrapper {
       sender,
     );
   }
-
+  async getSeq() {
+    return Number(this.wallet.account.sequence);
+  }
   /**
    * submitSoftwareUpgradeProposal creates proposal.
    */
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -274,7 +274,6 @@ export class CosmosWrapper {
       }
     }
     error = error ?? new Error('failed to submit tx');
-    console.log(error);
     throw error;
   }
 
@@ -1233,7 +1232,7 @@ export const mnemonicToWallet = async (
   }
   return new Wallet(address, account, pubKey, privKey, addrPrefix);
 };
-export const createAddress = async (mnemonicQA?: string) => {
+export const createAddress = async (mnemonicQA: string) => {
   const privKey = new proto.cosmos.crypto.secp256k1.PrivKey({
     key: await cosmosclient.generatePrivKeyFromMnemonic(mnemonicQA),
   });
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -192,9 +192,6 @@ cosmosclient.codec.register(
 );
 
 export class CosmosWrapper {
-  sendTokensWithRetry(cm: CosmosWrapper, address: string, arg2: string, sequence: number) {
-    throw new Error('Method not implemented.');
-  }
   sdk: cosmosclient.CosmosSDK;
   blockWaiter: BlockWaiter;
   wallet: Wallet;
--- a/src/helpers/cosmos.ts
+++ b/src/helpers/cosmos.ts
@@ -217,6 +217,7 @@ export class CosmosWrapper {
     msgs: T[],
     numAttempts = 10,
     mode: rest.tx.BroadcastTxMode = rest.tx.BroadcastTxMode.Async,
+    sequence: number = this.wallet.account.sequence,
   ): Promise<CosmosTxV1beta1GetTxResponse> {
     const protoMsgs: Array<google.protobuf.IAny> = [];
     msgs.forEach((msg) => {
@@ -234,7 +235,7 @@ export class CosmosWrapper {
               mode: proto.cosmos.tx.signing.v1beta1.SignMode.SIGN_MODE_DIRECT,
             },
           },
-          sequence: this.wallet.account.sequence,
+          sequence: sequence,
         },
       ],
       fee,
@@ -273,6 +274,7 @@ export class CosmosWrapper {
       }
     }
     error = error ?? new Error('failed to submit tx');
+    console.log(error);
     throw error;
   }
 
@@ -406,13 +408,20 @@ export class CosmosWrapper {
       gas_limit: Long.fromString('200000'),
       amount: [{ denom: this.denom, amount: '1000' }],
     },
+    sequence: number = this.wallet.account.sequence,
   ): Promise<InlineResponse20075TxResponse> {
     const msgSend = new proto.cosmos.bank.v1beta1.MsgSend({
       from_address: this.wallet.address.toString(),
       to_address: to,
       amount: [{ denom: this.denom, amount }],
     });
-    const res = await this.execTx(fee, [msgSend]);
+    const res = await this.execTx(
+      fee,
+      [msgSend],
+      10,
+      rest.tx.BroadcastTxMode.Block,
+      sequence,
+    );
     return res?.tx_response;
   }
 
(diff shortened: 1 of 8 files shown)


albertandrejev:
Please add JIRA task number to the PR description

MrTalecky:
> Please add JIRA task number to the PR description

Done

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +112,111 @@ export class TestStateLocalCosmosTestNet {
...
+    const mnemonicQA = generateMnemonic();
+    const mnemonicQATwo = generateMnemonic();
+    await this.createQaWallet(mnemonicQA, neutronPrefix);
+    await this.createQaWalletTwo(mnemonicQATwo, cosmosPrefix);
 
     this.wallets = {};
-    this.wallets.neutron = await walletSet(this.sdk1, neutron_prefix);
-    this.wallets.cosmos = await walletSet(this.sdk2, cosmos_prefix);
+    this.wallets.neutron = await walletSet(this.sdk1, neutronPrefix);
+    this.wallets.cosmos = await walletSet(this.sdk2, cosmosPrefix);
```
albertandrejev, line 122: I think you can create these wallets before `createQaWallet` method call. In this case you can avoid creation of the `tmpWallet`, and can simply use same wallet from `this.wallets`, or pass appropriate wallet to `createQaWallet`. This will simplify `createQaWallet` method's call.

src/helpers/cosmos.ts
```diff
@@ -708,7 +711,9 @@ export class CosmosWrapper {
       sender,
     );
   }
-
+  async getSeq() {
```
MrTalecky, line 714: Getting the current sequence number

src/testcases/common_localcosmosnet.ts
```diff
@@ -137,6 +138,40 @@ export class TestStateLocalCosmosTestNet {
       mnemonicQATwo,
     );
   };
+  sendTokensWithRetry = async (
```
MrTalecky, line 138: In "`sendTokensWithRetry`" We then enter a loop and attempt to call msgSend with the current sequence. If the call succeeds, we break out of the loop.
If it fails, we check if the error message includes the string "_incorrect account sequence_", which indicates that the sequence number has been used before. If it does, we fetch the current sequence number and increment it by one, and set the sequence variable to the new value. If the error message does not include "i_ncorrect account sequence_"", we simply increment the sequence variable by one. We then retry the `msgSend` call with the new sequence number.

src/helpers/cosmos.ts
```diff
@@ -273,6 +274,7 @@ export class CosmosWrapper {
       }
     }
     error = error ?? new Error('failed to submit tx');
+    console.log(error);
```
albertandrejev, line 277: Do not forget to remove this console output please.

src/helpers/cosmos.ts
```diff
@@ -1203,7 +1212,14 @@ export const mnemonicToWallet = async (
   }
   return new Wallet(address, account, pubKey, privKey, addrPrefix);
 };
-
+export const createAddress = async (mnemonicQA?: string) => {
```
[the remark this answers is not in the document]
albertandrejev: Please remove `?` from parameter

src/testcases/common_localcosmosnet.ts
```diff
@@ -53,6 +60,18 @@ const walletSet = async (
...
+  prefix: string,
+  mnemonicQA: string,
+): Promise<Record<string, Wallet>> => ({
+  genQaWal1: await mnemonicToWallet(
+    cosmosclient.AccAddress,
+    sdk,
+    mnemonicQA,
+    prefix,
+  ),
+});
```
albertandrejev, line 74: You can simplify this function
```suggestion
const walletSetQa = async (
  sdk: cosmosclient.CosmosSDK,
  prefix: string,
  mnemonicQA: string,
): Promise<Record<string, Wallet>> => await mnemonicToWallet(
    cosmosclient.AccAddress,
    sdk,
    mnemonicQA,
    prefix,
  );
```

src/helpers/cosmos.ts
albertandrejev: No, you need to get sequence from blockchain, and not from wallet.
try to use 
```
const account = await rest.auth
      .account(sdk, address)
      .then((res) =>
        cosmosclient.codec.protoJSONToInstance(
          cosmosclient.codec.castProtoJSONOfProtoAny(res.data.account),
        ),
      )
      .catch((e) => {
        console.log(e);
        throw e;
      });

    if (!(account instanceof proto.cosmos.auth.v1beta1.BaseAccount)) {
      throw new Error("can't get account");
    }
```
and get sequence from account. In this case you will get actual sequence id

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +100,101 @@ export class TestStateLocalCosmosTestNet {
...
+      amount: [{ denom: cm.denom, amount: '1000' }],
+    };
+    let sequence = startingSequence;
+    let retryCount = 0;
+    while (true) {
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
+        break;
+      } catch (e) {
+        if (retryCount < 4) {
```
albertandrejev, line 155: you can get rid of this `if`, and move condition check into `while (retryCount < 4)`

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +100,101 @@ export class TestStateLocalCosmosTestNet {
...
+          if (e.response && e.response.data && e.response.data.error) {
+            const errorMsg = e.response.data.error;
+            const sequenceErrMsg = 'incorrect account sequence';
+            if (errorMsg.includes(sequenceErrMsg)) {
+              await cm.blockWaiter.waitBlocks(1);
+              sequence = Number(await cm.getSeq());
+            }
+          }
```
albertandrejev, line 163: You can simplify this and simply use new sequence id each time without checking for error message, and reveal error message only after all retries will be exhausted
something like
```
...
catch (e) {
  sequence = newSeqId();
  retryCount++;
}
```

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +100,101 @@ export class TestStateLocalCosmosTestNet {
...
+    const fee = {
+      gas_limit: Long.fromString('200000'),
+      amount: [{ denom: cm.denom, amount: '1000' }],
+    };
+    let sequence = startingSequence;
+    let retryCount = 0;
+    while (true) {
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
+        break;
```
albertandrejev, line 153: better use `return` here.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +100,101 @@ export class TestStateLocalCosmosTestNet {
...
+            const sequenceErrMsg = 'incorrect account sequence';
+            if (errorMsg.includes(sequenceErrMsg)) {
+              await cm.blockWaiter.waitBlocks(1);
+              sequence = Number(await cm.getSeq());
+            }
+          }
+          retryCount++;
+          sequence = Number((await cm.getSeq()) + 1);
+        } else {
+          throw new Error('Failed to send tokens after 4 retries');
```
albertandrejev, line 167: move this out of `while` body.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +100,101 @@ export class TestStateLocalCosmosTestNet {
...
+      valAddr: `${prefix}valoper`,
+      valPub: `${prefix}valoperpub`,
+      consAddr: `${prefix}valcons`,
+      consPub: `${prefix}valconspub`,
+    });
+
+    const address = await createAddress(mnemonic);
+    const sequence = Number(await cm.getSeq());
+    await this.sendTokensWithRetry(cm, address, '5500000000', sequence);
+    const balances = await cm.queryBalances(address);
```
albertandrejev, line 195: I'd recommend you to wait for 1 block. Or even better use `getWithAttempts` function.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +100,101 @@ export class TestStateLocalCosmosTestNet {
...
+    await this.createQaWallet(
+      mnemonicQATwo,
+      cosmosPrefix,
+      this.sdk2,
+      this.blockWaiter2,
+      this.wallets.cosmos.demo2,
+      COSMOS_DENOM,
+    );
+
+    this.wallets.qaOne = await walletSetQa(
```
albertandrejev, line 127: If you will refactor `walletSetQa` to return wallet only without any additional fields, then in this case you can call it like:
```
this.wallets.qaOne = { genQaWal1: await walletSetQa(.....) }
```

src/helpers/cosmos.ts
MrTalecky: That's what I was talking about. Which is probably the wrong way to ask for sequence.
Thank you.

src/helpers/cosmos.ts
```diff
@@ -708,8 +708,27 @@ export class CosmosWrapper {
       sender,
     );
   }
-  async getSeq() {
-    return Number(this.wallet.account.sequence);
+  async getSeq(
```
MrTalecky, line 711: This function returns the actual sequence from the blockchain

src/testcases/common_localcosmosnet.ts
MrTalecky: @albertandrejev If we put the condition check in while, we will have to do 4 queries every time. 
But in some cases we don't need to make 4 tries, and it is enough to send once.

Or am I wrong? 

Like this:
```
 while (retryCount < 4) {
      try {
        await cm.msgSend(to, amount, fee, sequence);
        retryCount++;
        return;
      } catch (e) { ...
```

albertandrejev: no, but you need to increment in the `catch` section, and if everything is fine then return in the `try` section will stop the loop

src/helpers/cosmos.ts
albertandrejev: and yes and no, because it returns sequence that might be outdated because it contains data on moment of the wallet creation (when you are adding it to the tests).

src/testcases/common_localcosmosnet.ts
```diff
@@ -196,7 +188,7 @@ export class TestStateLocalCosmosTestNet {
 
     const address = await createAddress(mnemonic);
     const sequence = await cm.getSeq(sdk, walletAddress);
-    console.log(sequence);
+    await blockWaiter.waitBlocks(1);
```
MrTalecky, line 191: I added Blockwaiter here, for now.

src/helpers/cosmos.ts
MrTalecky: Wait, what's the best way to implement it then? 
It's not like the wallet address changes at the time of creation or afterwards. 
The `getSeq` function gives based on _two_ arguments, one of which is the wallet address, the account sequence number.

albertandrejev: You need to request for new sequence from network instead of incrementing it. See this comment https://github.com/neutron-org/neutron-integration-tests/pull/81#discussion_r1108878911
And if you are requesting data from network using then it should return real sequence

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +102,147 @@ export class TestStateLocalCosmosTestNet {
...
+      mnemonicQAFive,
+      neutronPrefix,
+      this.sdk1,
+      this.blockWaiter1,
+      this.wallets.neutron.demo1,
+      NEUTRON_DENOM,
+      this.wallets.neutron.demo1.address,
+    );
+
+    this.wallets.qaNeutron = await walletSetQa(
```
albertandrejev, line 161: I think it is better to call `walletSetQa` inside `this.createQaWallet` and return wallet from it. It will make code better to read and group similar operations together.

albertandrejev: In this case you can completely remove `walletSetQa` function and use `mnemonicToWallet` instead of it.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +102,147 @@ export class TestStateLocalCosmosTestNet {
...
+    startingSequence: number,
+  ) => {
+    const fee = {
+      gas_limit: Long.fromString('200000'),
+      amount: [{ denom: cm.denom, amount: '1000' }],
+    };
+    let sequence = startingSequence;
+    let retryCount = 0;
+
+    while (retryCount < 4) {
```
albertandrejev, line 200: move number into function parameter with default value, and set it to something like 10.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +102,147 @@ export class TestStateLocalCosmosTestNet {
...
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
```
albertandrejev, line 202: you can add `cons sequence = await cm.getSeq(cm.sdk, cm.wallet.address);` before this line. After this you can remove `startingSequence: number` parameter from method signature.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +102,147 @@ export class TestStateLocalCosmosTestNet {
...
+        return;
+      } catch (e) {
+        await cm.blockWaiter.waitBlocks(1);
+        retryCount++;
+        sequence = await cm.getSeq(cm.sdk, cm.wallet.address);
+        if (retryCount === 4) {
+          throw new Error(`Failed to send tokens after ${retryCount} retries.`);
+        }
```
albertandrejev, line 210: you can completely remove this code. Simply move `throw new Error(`Failed to send tokens after ${retryCount} retries.`)` after this `while` loop, and it will be called after loop will finish without any success.

src/testcases/common_localcosmosnet.ts
albertandrejev: Also in this case you can get rid of mnemonic creation in this method. So you can create everything in one method. This will reduce code duplication.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +102,147 @@ export class TestStateLocalCosmosTestNet {
...
+    let retryCount = 0;
+
+    while (retryCount < 4) {
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
+        return;
+      } catch (e) {
+        await cm.blockWaiter.waitBlocks(1);
+        retryCount++;
+        sequence = await cm.getSeq(cm.sdk, cm.wallet.address);
```
albertandrejev, line 207: If you will move this line before `msgSend` then you can get rid of it here.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +102,147 @@ export class TestStateLocalCosmosTestNet {
...
+
+    const address = await createAddress(mnemonic);
+    const sequence = await cm.getSeq(sdk, walletAddress);
+    await this.sendTokensWithRetry(
+      cm,
+      toString(address),
+      '10500000000',
+      sequence,
+    );
+    const balances = await cm.queryBalances(toString(address));
```
albertandrejev, line 243: I think it is better to check for balance inside `sendTokensWithRetry`. Because this is responsibility of this method.

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +90,121 @@ export class TestStateLocalCosmosTestNet {
     );
 
     await setup(host1, host2);
+    const mnemonicQA = generateMnemonic();
+    const mnemonicQATwo = generateMnemonic();
+    const mnemonicQAThree = generateMnemonic();
+    const mnemonicQAFour = generateMnemonic();
+    const mnemonicQAFive = generateMnemonic();
```
ratik, line 97: as we using these vars only once you can remove them and use like 
```
    this.wallets.qaCosmos = await this.createQaWallet(
      generateMnemonic(),
      cosmosPrefix,
      this.sdk2,
      this.blockWaiter2,
      this.wallets.cosmos.demo2,
      COSMOS_DENOM,
    );
```

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +90,121 @@ export class TestStateLocalCosmosTestNet {
...
+      this.wallets.neutron.demo1,
+      NEUTRON_DENOM,
+    );
+  };
+  sendTokensWithRetry = async (
+    cm: CosmosWrapper,
+    to: string,
+    amount: string,
+    address: AccAddress,
+  ) => {
```
ratik, line 149:
```suggestion
  ): Promise<void> => {
```

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +90,121 @@ export class TestStateLocalCosmosTestNet {
...
+      amount: [{ denom: cm.denom, amount: '1000' }],
+    };
+    let sequence = await cm.getSeq(cm.sdk, cm.wallet.address);
+    let retryCount = 0;
+
+    while (retryCount < 10) {
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
+        return;
+      } catch (e) {
```
ratik, line 161: mb we can check `e` here and if there is sequence error -> retry, if other -> re-throw `e`

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +90,121 @@ export class TestStateLocalCosmosTestNet {
...
+      valPub: `${prefix}valoperpub`,
+      consAddr: `${prefix}valcons`,
+      consPub: `${prefix}valconspub`,
+    });
+
+    const address = await createAddress(mnemonic);
+    await this.sendTokensWithRetry(
+      cm,
+      toString(address),
+      '10500000000',
```
ratik, line 197: move amount into params and set default value to `10500000000`

src/testcases/common_localcosmosnet.ts
```diff
@@ -81,9 +90,121 @@ export class TestStateLocalCosmosTestNet {
...
+    }
+    const balances = await cm.queryBalances(toString(address));
+    if (balances == null) {
+      throw new Error('Could not put tokens on the generated wallet.');
+    }
+    throw new Error(`Failed to send tokens after ${retryCount} retries.`);
+  };
+
+  createQaWallet = async (
+    mnemonic: string,
```
ratik, line 175: mb we can even remove mnemonic from params and generate in fn body

PR #82, opened 2023-02-13
Update Makefile to include neutron-tests CI logic

keylogic1:
This PR only adds rules to the `Makefile` without breaking existing ones

This will allow us to use this repo's build information from now on without relying on `neutron-tests`


oopcode:
what do you think about adding a new Makefile variable, `APP_DIR`, with the default value equal to the ones that we use on our local machines? that would allow to reduce code duplication, for example

(if the ci build was substantially different from a local one, introducing a new command would make sense; but here it's jsut a matter of configuration, it seems)

keylogic1:
@oopcode Thanks, indeed your proposal is better. I've updated the Makefile to reflect this

oopcode (approved):

oldremez (approved):

PR #86, opened 2023-03-01
feat: setup and teardown imrovement

ratik:
Moved setup/teardown to jest params, so it will work well in parallel tests

https://github.com/neutron-org/neutron-tests/actions/runs/4308061172


quasisamurai (commented):
LGTM, waiting for tests run

oopcode (approved):

PR #87, opened 2023-03-01
Feat/ntrn 312 run parallel tests

MrTalecky:
**This branch contains all the changes**:
* several features have been added that generate new wallets
* new wallets are put into tests, which can be run in parallel
* test suites are divided into 2 groups: `parallel` and `run in band`.
* also added corrections for all comments from the branch `for_parallel_tests`
* 
**Open questions without which we can't approve PR**:
* the first test in the `simple` suite is probably pointless. It checks for hardcoded wallet addresses.
* Sometimes several tests in the `treasury` suite crash. This happens when running in parallel 1 in 5 times. I haven't been able to find the cause yet, maybe in the review process it will be clearer.
* I experimented and didn't understand the reason for periodic test crashes in `interchain_tx_query`. Some test crashes are accompanied by an `account sequence mismatch` error. I tested the hypothesis that the sequence is taken from one wallet (default parameter) and used in the transaction another, but the same wallet is used there.

Link to successful test run: https://github.com/neutron-org/neutron-tests/actions/runs/4384105049


src/testcases/common_localcosmosnet.ts
```diff
@@ -83,7 +92,107 @@ export class TestStateLocalCosmosTestNet {
...
+    );
+    this.wallets.qaNeutronFive = await this.createQaWallet(
+      neutronPrefix,
+      this.sdk1,
+      this.blockWaiter1,
+      this.wallets.neutron.demo1,
+      NEUTRON_DENOM,
+    );
+  };
+  sendTokensWithRetry = async (
```
albertandrejev, line 133: Add empty line here please.

src/testcases/common_localcosmosnet.ts
```diff
@@ -83,7 +92,107 @@ export class TestStateLocalCosmosTestNet {
...
+    let sequence = await cm.getSeq(cm.sdk, cm.wallet.address);
+
+    while (retryCount > 0) {
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
+        return;
+      } catch (e) {
+        if (e.message.includes('sequence')) {
+          await cm.blockWaiter.waitBlocks(1);
+          retryCount--;
```
albertandrejev, line 153: It is not good idea to change parameter variable value in the function body. Better compare `while` condition against this value and your internal variable.

src/testcases/common_localcosmosnet.ts
```diff
@@ -83,7 +92,107 @@ export class TestStateLocalCosmosTestNet {
...
+    to: string,
+    amount: string,
+    address: AccAddress,
+    retryCount = 10,
+  ): Promise<void> => {
+    const fee = {
+      gas_limit: Long.fromString('200000'),
+      amount: [{ denom: cm.denom, amount: '1000' }],
+    };
+    let sequence = await cm.getSeq(cm.sdk, cm.wallet.address);
```
albertandrejev, line 144: make this line `const` and move it above line 148 (right before `cm.msSend`)

src/testcases/common_localcosmosnet.ts
```diff
@@ -83,7 +92,107 @@ export class TestStateLocalCosmosTestNet {
...
+
+    while (retryCount > 0) {
+      try {
+        await cm.msgSend(to, amount, fee, sequence);
+        return;
+      } catch (e) {
+        if (e.message.includes('sequence')) {
+          await cm.blockWaiter.waitBlocks(1);
+          retryCount--;
+          sequence = await cm.getSeq(cm.sdk, cm.wallet.address);
```
albertandrejev, line 154: remove this line

src/testcases/common_localcosmosnet.ts
```diff
@@ -83,7 +92,107 @@ export class TestStateLocalCosmosTestNet {
...
+        } else {
+          throw e;
+        }
+      }
+    }
+    const balances = await cm.queryBalances(toString(address));
```
albertandrejev, line 160: why not `address.toString()` ?

src/testcases/common_localcosmosnet.ts
```diff
@@ -1,9 +1,18 @@
...
+  COSMOS_DENOM,
+  createAddress,
+  mnemonicToWallet,
+} from '../helpers/cosmos';
 import { setup } from '../helpers/env';
 import { BlockWaiter } from '../helpers/wait';
+import { generateMnemonic } from 'bip39';
+import { CosmosWrapper, NEUTRON_DENOM } from '../helpers/cosmos';
+import Long from 'long';
+import { toString } from 'lodash';
```
albertandrejev, line 14: Are you sure that we need `lodash` package here at all?

src/testcases/common_localcosmosnet.ts
MrTalecky: Corrected)

src/testcases/common_localcosmosnet.ts
MrTalecky: At one point I took toString out of it, yes, that's right, we don't need to.

src/testcases/common_localcosmosnet.ts
MrTalecky: The same here, btw, maybe it's should change
https://github.com/neutron-org/neutron-integration-tests/blob/main/src/helpers/cosmos.ts#L218

src/testcases/common_localcosmosnet.ts
```diff
@@ -83,7 +91,108 @@ export class TestStateLocalCosmosTestNet {
...
+      } catch (e) {
+        if (e.message.includes('sequence')) {
+          await cm.blockWaiter.waitBlocks(1);
+          attemptCount++;
+        } else {
+          throw e;
+        }
+      }
+    }
+    const balances = await cm.queryBalances(address.toString());
```
albertandrejev, line 160: code will reach this point only after `retryCount` and only if `msgSend` will not be succeed. Are you sure that checking balance at this point is required?

MrTalecky: Yes, indeed, this test is irrelevant here. I'll move it.

package.json
```diff
@@ -4,17 +4,19 @@
...
-    "test:interchaintx": "jest --runInBand -b src/testcases/interchaintx",
-    "test:simple": "jest --runInBand -b src/testcases/simple",
-    "test:interchain_kv_query": "jest --runInBand -b src/testcases/interchain_kv_query",
-    "test:interchain_tx_query": "jest --runInBand -b src/testcases/interchain_tx_query_plain",
+    "test": "yarn test:parallel && yarn test:run_in_band",
+    "test:parallel": "jest -b src/testcases/parallel",
+    "test:run_in_band": "yarn test:governance && yarn test:interchaintx && yarn test:interchain_kv_query && yarn test:interchain_tx_query_plain && yarn test:subdao && yarn test:treasury",
+    "test:simple": "jest -b src/testcases/parallel/simple",
+    "test:interchaintx": "jest -b src/testcases/run_in_band/interchaintx",
+    "test:interchain_kv_query": "jest -b src/testcases/run_in_band/interchain_kv_query",
```
ratik, line 12: remove it please. `plain` is copy of it

ratik (approved):

src/testcases/common_localcosmosnet.ts
```diff
@@ -80,7 +84,129 @@ export class TestStateLocalCosmosTestNet {
...
+    const mnemonic = generateMnemonic();
+    const newWallet = await mnemonicToWallet(
+      cosmosclient.AccAddress,
+      sdk,
+      mnemonic,
+      prefix,
+      false,
+    );
+    await this.sendTokensWithRetry(cm, newWallet.address, tokens);
+    const w = await mnemonicToWallet(
```
albertandrejev, line 203: please change this variable name

src/helpers/cosmos.ts
```diff
@@ -1092,8 +1116,9 @@ export const mnemonicToWallet = async (
     consAddr: `${addrPrefix}valcons`,
     consPub: `${addrPrefix}valconspub`,
   });
+  const address = walletType.fromPublicKey(pubKey);
```
albertandrejev, line 1119: Please move all code related to address generation to the separate function. After that you can get rid of `validate` variable, and generate address from mnemonic in any place where you need it (see other comment).

src/testcases/common_localcosmosnet.ts
```diff
@@ -80,7 +84,129 @@ export class TestStateLocalCosmosTestNet {
...
+  ) {
+    const cm = new CosmosWrapper(sdk, blockWaiter, wallet, denom);
+    const mnemonic = generateMnemonic();
+    const newWallet = await mnemonicToWallet(
+      cosmosclient.AccAddress,
+      sdk,
+      mnemonic,
+      prefix,
+      false,
+    );
```
albertandrejev, line 201: If you will move address generation from mnemonic to separate function then you can use it here (see my other comment).

src/helpers/cosmos.ts
albertandrejev: ok, let's leave it as is. I just don't like to have selector variable to switch code behavior based on it. But ok.

albertandrejev (approved):

src/helpers/cosmos.ts
MrTalecky: Yes, we can come back to this as part of the task of refactoring, for example, or correct it when we divide the sequential tests into parallel ones.

PR #95, opened 2023-03-10
Remove hard-coded contract addresses from tests

oldremez:
This PR gets rid of all DAO-related hardcoded addresses in tests.
1. The core DAO address is queried from admin module
2. Most other contracts (voting modules, proposals, and so on) are queried recursively from the DAO core
3. The treasury is queried from feeburner module

This PR already has the information about overrule proposals, so it works only with this PR to Neutron: https://github.com/neutron-org/neutron/pull/161

Tests run: https://github.com/neutron-org/neutron-tests/actions/runs/4393949558


src/testcases/run_in_band/interchain_kv_query.test.ts
```diff
@@ -161,8 +158,11 @@ const acceptInterchainqueriesParamsChangeProposal = async (
   value: string,
   amount = '1000',
 ) => {
+  const daoCoreAddress = (await cm.getChainAdmins())[0];
+  const daoContracts = await getDaoContracts(cm, daoCoreAddress);
+
   const proposalTx = await cm.submitParameterChangeProposal(
-    PRE_PROPOSE_CONTRACT_ADDRESS,
+    daoContracts.proposal_modules.single.pre_proposal_module.address,
```
sotnikov-s, line 165: please add nice short variables for contract addresses just as you've done in gov tests

src/testcases/run_in_band/subdao.test.ts
```diff
@@ -41,42 +40,52 @@ describe('Neutron / Subdao', () => {
...
       main_dao_addr.toString(),
       security_dao_addr.toString(),
     );
 
-    await cm.bondFunds(VAULT_CONTRACT_ADDRESS, '10000');
+    const voting_module_address = await cm_main_dao.queryContract<string>(
+      subDAO.core.address,
+      { voting_module: {} },
+    );
+    const voting_vaults = await cm_main_dao.queryContract<
```
sotnikov-s, line 65: here you use the same flow of getting vaults of a dao as you do in the `dao.ts` `getDaoContracts`. would be nice to move this part to a func, say, `getDaoVotingVaults` which would return for example this:
```ts
voting_vaults: {
        ntrn_vault: { address: ntrn_vault_address },
        lockdrop_vault: { address: lockdrop_vault_address },
      }
```

and reuse it here and in the `getDaoContracts`

src/testcases/run_in_band/subdao.test.ts
```diff
@@ -816,20 +914,33 @@ const setupSubDaoTimelockSet = async (
   const timelockCodeId = parseInt(
     await cm.storeWasm(NeutronContract.SUBDAO_TIMELOCK),
   );
+
+  const mainDaoCore = (await cm.getChainAdmins())[0];
+  const daoContracts = await getDaoContracts(cm, mainDaoCore);
+
+  const votingModuleContractInfo = await cm.getContractInfo(
+    daoContracts['voting_module'].address,
```
sotnikov-s, line 922: why `daoContracts['voting_module']`, not `daoContracts.voting_module`? same below

src/helpers/cosmos.ts
```diff
@@ -595,8 +586,12 @@ export class CosmosWrapper {
     sender: string,
     options: MultiChoiceOption[],
   ): Promise<InlineResponse20075TxResponse> {
+    const daoCoreAddress = (await this.getChainAdmins())[0];
+    const daoContracts = await getDaoContracts(this, daoCoreAddress);
```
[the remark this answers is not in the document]
oldremez: Yeah, I thought about it but haven't done it. After some consideration, I decided not to do that since we might use the `getDaoContracts` function not only for the DAO that is the chain admin. E.g. I kinda advocate for deploying separate DAO in tests (where it isn't necessary to execute admin proposals) for the sake of parallelism and so on. Also, the function can be generalized to be used with subDAOs.

src/testcases/parallel/tokenomics.test.ts
```diff
@@ -5,15 +5,16 @@ import {
...
 } from '../../helpers/cosmos';
 import Long from 'long';
 import { getWithAttempts } from '../../helpers/wait';
+import { getTreasuryContract } from '../../helpers/dao';
 
 describe('Neutron / Tokenomics', () => {
   let testState: TestStateLocalCosmosTestNet;
   let cmNeutron: CosmosWrapper;
   let cmGaia: CosmosWrapper;
+  let treasuryContractAddress;
```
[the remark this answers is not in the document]
oldremez: yup, fixed

src/testcases/run_in_band/interchain_kv_query.test.ts
oldremez: done

src/testcases/run_in_band/subdao.test.ts
oldremez: fixed

src/helpers/cosmos.ts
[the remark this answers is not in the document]
sotnikov-s: subDAOs can vary in structure, e.g. proposals and voting vaults, whereas in this func there is a struct response relative to the core DAO only. this is why I suggested this. but the first point is fair, thus I resolve this conversation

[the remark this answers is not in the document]
oldremez: Yeah, I agree, that's why I said it should be "generalized" for it :)

src/testcases/run_in_band/subdao.test.ts
oldremez: Ok after some considerations actually I think I'll do it

sotnikov-s (approved):

oopcode (approved):

PR #98, opened 2023-03-12
feat: instantiate2

ratik:
So I'm against the idea about putting Token Generation Event (TGE) contracts into genesis:
1. It's pain to debug init scripts
2. It's almost impossible to predict start of `cosmopark` so you'll have to set really loose periods for contracts phases so it will take really long to run test
3. You'll have to maintain two repos instead of one
4. You can't deploy several sets of contracts
5. <del>It's ugly</del>
There is [Instantiate2 msg](https://docs.rs/cosmwasm-std/latest/cosmwasm_std/enum.WasmMsg.html#variant.Instantiate2) which can help us avoid it. 
I've added the proto into this PR and added this Msg into codec so now it can be used without replacing cosmos-client js lib

### DIFF
--- a/gen-proto.sh
+++ b/gen-proto.sh
@@ -10,6 +10,8 @@ for dir in $proto_dirs; do
   proto_files=("${proto_files[@]} $(find "${dir}" -maxdepth 1 -name '*.proto')")
 done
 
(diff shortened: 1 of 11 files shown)


ratik:
well. it seems address depends on instantiate msg which is ridiculous

PR #131, opened 2023-04-07
Testing hardened overrules #NTRN-330 #NTRN-325

oldremez:
This PR tests changes in overrules.

Besides testing overrules, this PR also addresses some other things:
1. it restores launching of ibc hooks tests (they were skipped by mistake)
2. it changes `CodeId` type to `number` across the tests
3. it removes `sender` argument from `executeContract` method since it can't be set to an address that differs from WalletWrapper's address
4. it adds more methods to work with DAO-related stuff (e.g. DAO deployment and getting of subdao contracts)
5. minor refactoring

Overrules changes: https://github.com/neutron-org/neutron-dao/pull/32
Genesis changes: https://github.com/neutron-org/neutron/pull/201


oldremez:
https://github.com/neutron-org/neutron-tests/actions/runs/4694337259

src/helpers/cosmos.ts
```diff
@@ -433,16 +457,16 @@ export class WalletWrapper {
       'code_id',
     ]);
 
-    return attributes[0].code_id;
+    return parseInt(attributes[0].code_id);
```
ratik, line 460:
```suggestion
    const codeId = parseInt(attributes[0].code_id, 10);
    if (isNaN(codeId)) {
       throw new Error('shit, codeid is fucked up');
    }
    return codeId;
```
but I'd prefer move changes like that into separate PR as out of scope.

oldremez:
https://github.com/neutron-org/neutron-tests/actions/runs/4699297637

src/helpers/cosmos.ts
oldremez: Yup 100% agree both on changes proposed and moving them as out of scope :)

oopcode (approved):

PR #132, opened 2023-04-07
download hermes binary instead of building it

oldremez:
Currently to run the integration tests locally one has to build the hermes image which implies building the hermes from sources. It takes significant time (like, dozens of minutes) while hermes version doesn't change much and it's not necessary to build it from scratch since it's the test environment.
This PR is to replace the building process with binary download. The caveat is if we will need a specific commit we'd need to roll back it to building once again.


oldremez:
https://github.com/neutron-org/neutron-tests/actions/runs/4670844556/jobs/8271920084

quasisamurai (approved):

PR #145, opened 2023-04-28
fix: clean staking error #ntrn-443

NeverHappened:


foxpy:
https://github.com/neutron-org/neutron-tests/actions/runs/5078714499

PR #158, opened 2023-05-08
Fix race in governance test when running in parallel

foxpy:
The race happens when DAO account receives NTRN tokens during governance test. Instead of checking for amount of NTRN tokens on the account of DAO, we simply check that send transaction succeeds. Also, when running governance test independently, we don't have to wait for IBC connections and such, so I have disabled these barriers for this test.


foxpy:
https://github.com/neutron-org/neutron-tests/actions/runs/4915602773

sotnikov-s (approved):

oopcode (approved):

PR #174, opened 2023-07-04
feat: support cosmos-sdk v0.47

swelf19:


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/7141161284

PR #175, opened 2023-07-07
feat: add stargate_queries test suite #ntrn-6

NeverHappened:


pr0n00gler (approved):

src/testcases/parallel/stargate_queries.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    });
+
+    test('interchainqueries params should work', async () => {
+      const res = JSON.parse(
+        await querySmart({ interchainqueries_params: {} }),
+      );
+      expect(+res.params.query_submit_timeout).toBeGreaterThan(0);
+    });
+
+    test('feeburner params should work', async () => {
```
swelf19, line 265: any query that should not wotk?

NeverHappened: well, all the other existing not whitelisted stargate queries :)
I can add test for one of them to show that, but dont think it has much value

swelf19: it has some value, to show we have a whitelist, not just whole stargate queries set enabled.

swelf19 (approved):

src/testcases/parallel/stargate_queries.test.ts
NeverHappened: seems reasonable! will add a query to test that

PR #177, opened 2023-07-10
Feat: NTRN-29. Add global fee module

albertandrejev:
Test for the globalfees module on the Netron.

https://hadronlabs.atlassian.net/browse/NTRN-29

Related PRs:
https://github.com/neutron-org/neutron/pull/264
https://github.com/neutron-org/neutron-docs/pull/91

### DIFF
--- a/src/testcases/run_in_band/globalfee.test.ts
+++ b/src/testcases/run_in_band/globalfee.test.ts
@@ -63,65 +63,65 @@ describe('Neutron / Global Fee', () => {
     ]);
   });
 
-  test('change minimum gas price parameter', async () => {
-    const proposalId = await daoMember.submitParameterChangeProposal(
-      'Minimumm Gas Price Change Proposal',
-      'Param change proposal. It will change the minimum gas price of the global fee module.',
-      'globalfee',
-      'MinimumGasPricesParam',
-      '[{"denom": "untrn", "amount": "0.01"}]',
-      '1000',
-    );
+  // test('change minimum gas price parameter', async () => {
+  //   const proposalId = await daoMember.submitParameterChangeProposal(
+  //     'Minimumm Gas Price Change Proposal',
+  //     'Param change proposal. It will change the minimum gas price of the global fee module.',
+  //     'globalfee',
+  //     'MinimumGasPricesParam',
+  //     '[{"denom": "untrn", "amount": "0.01"}]',
+  //     '1000',
+  //   );
 
-    await daoMember.voteYes(proposalId);
-    await dao.checkPassedProposal(proposalId);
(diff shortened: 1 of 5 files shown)


pr0n00gler (approved):

foxpy (approved):

src/testcases/run_in_band/globalfee.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    neutronAccount = new WalletWrapper(
+      neutronChain,
+      testState.wallets.qaNeutron.genQaWal1,
+    );
+    const daoCoreAddress = (await neutronChain.getChainAdmins())[0];
+    const daoContracts = await getDaoContracts(neutronChain, daoCoreAddress);
+    dao = new Dao(neutronChain, daoContracts);
+    daoMember = new DaoMember(neutronAccount, dao);
+    await daoMember.bondFunds('1000');
+    await getWithAttempts(
```
sotnikov-s, line 36: what's the point of getting with attempts here? bondFunds uses executeContract inside it which is a sync op, so the bonded funds must become accrued voting power just after the bondFunds. getting with attempts in such case would only complicate code and increase time consumed by the test in a failing case here

src/testcases/run_in_band/globalfee.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      async () =>
+        await dao.queryVotingPower(daoMember.user.wallet.address.toString()),
+      async (response) => response.power == 1000,
+      20,
+    );
+
+    await daoMember.user.msgSend(dao.contracts.core.address, '1000', {
+      gas_limit: Long.fromString('200000'),
+      amount: [{ denom: daoMember.user.chain.denom, amount: '500' }],
+    });
```
sotnikov-s, line 47: why this send?

src/testcases/run_in_band/globalfee.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  afterAll(async () => {
+    await daoMember.unbondFunds('1000');
```
sotnikov-s, line 51: why unbond?

sotnikov-s (approved):

src/testcases/run_in_band/globalfee.test.ts
albertandrejev: It was added there to be fully sure that bond funds was successfully bonded t will throw exception if something goes wrong.

src/testcases/run_in_band/globalfee.test.ts
albertandrejev: It was taken from `governance.tsst.ts` test, and I forgot to remove it from current test.

PR #181, opened 2023-07-18
Chore. Upgrade hermes verision to 1.5.1

albertandrejev:


NeverHappened:
I've updated to 1.6.0 :D
https://github.com/neutron-org/neutron-integration-tests/pull/182
It has this new pull feature, so I wanted it right away

PR #223, opened 2023-10-16
add sdk-47 params test NTRN-113

quasisamurai:
[TASK](https://hadronlabs.atlassian.net/browse/NTRN-113)


quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/6533928970/job/17747811926

pr0n00gler (approved):

src/testcases/run_in_band/parameters.test.ts
```diff
@@ -0,0 +1,380 @@
...
+      const res = await daoMember1.user.msgSend(
+        dao.contracts.core.address,
+        '1000',
+      );
+      expect(res.code).toEqual(0);
+    });
+  });
+
+  describe('create several proposals', () => {
+    test('create proposal #1, will pass', async () => {
```
sotnikov-s, line 119: apparently you've used some parts from governance test code to create this one. there was a point in phrases like "create proposal **#1, will pass**", and "create several proposals", but in this test file it's rather confusing: it's not just several proposals, it's proposals serving concrete goal, and they all pass as far as I can see, we don't test passed/failed proposal logic here. so it would be much clean if there will only be descriptions related to this very test file, no more no less. please consider:
1. cleaning up the file from ambiguous descriptions
2. giving test suites more meaningful names e.g. `create proposal #1, will pass` -> `create param change proposal for interchainqueries`. same below where you execute proposals

sotnikov-s: additionally, since it's not a gov test, but params update test, another simplification that you can apply is reducing the number of dao members to 1. this would 1) shorten code 2) remove confusion around YES NO YES votes 3) reduce test run time

src/testcases/run_in_band/parameters.test.ts
```diff
@@ -0,0 +1,380 @@
...
+    test('execute passed proposal', async () => {
+      const host = await neutronChain.queryHostEnabled();
+      expect(host).toEqual(true);
+      await daoMember1.executeProposalWithAttempts(proposalId);
+    });
+    test('check if params changed after proposal execution', async () => {
+      const params = await neutronChain.queryInterchainqueriesParams();
+      expect(params.params.query_submit_timeout).toEqual('30');
+      expect(params.params.tx_query_removal_limit).toEqual('20');
+    });
```
sotnikov-s, line 211: this is kind of annoying — to compare results from this part of the code with the ones in the proposal above: one should look here, memorise values, scroll above, compare, and do all these 7 times for each module. would be much better to have test for a module placed as a whole between other modules' test cases, e.g.

icq module proposal created, voted for, executed, checked, then the same for tokenfactory, and so on

sotnikov-s: another idea is to move these new parameters to constants and reuse them in both proposal creation and in assertions

src/testcases/run_in_band/parameters.test.ts
```diff
@@ -0,0 +1,380 @@
...
+    test('check if proposal is passed', async () => {
+      await dao.checkPassedProposal(proposalId);
+    });
+    test('execute passed proposal', async () => {
+      await daoMember1.executeProposalWithAttempts(proposalId);
+    });
+    test('check if params changed after proposal execution', async () => {
+      const params = await neutronChain.queryTokenfactoryParams();
+      expect(params.params.denom_creation_fee).toHaveLength(0);
+      expect(params.params.denom_creation_gas_consume).toEqual('100000');
```
sotnikov-s, line 238: how do we make sure these new values are actually not the same as the old ones? e.g. if we ever change the neutron tokenfactory genesis to have `denom_creation_fee=0` and `denom_creation_gas_consume= 100000`, we won't notice a possible bug because this test won't work properly. please add checks all over the test file that new params are not the same as old params

sotnikov-s (changes requested):

src/testcases/run_in_band/parameters.test.ts
NeverHappened: Agree with above - how about really to create tests, in which each test would go through complete cycle?
Link this:
- save previous param values
- create proposal
- vote for it (can be one dao member as mentioned above)
- execute it
- check new param values are different from the old ones

NeverHappened (changes requested):

NeverHappened (approved):

src/testcases/run_in_band/parameters.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      );
+    });
+    test('check voting power', async () => {
+      await getWithAttempts(
+        neutronChain.blockWaiter,
+        async () => await dao.queryTotalVotingPower(),
+        async (response) => response.power == 1000,
+        20,
+      );
+    });
```
sotnikov-s, line 53: what's the point of using a complicated func like getWithAttempts here? it makes sense to use it in async/await flows like interchain transactions where we don't know for sure when the result comes. here it's enough to just wait a block and make a simple query — results should be available straight after the pause. yeah it doesn't really matter here in terms of pros but a most obvious con here is the error case: if there's an error, we won't make 20 tries waiting for a success result, but fail immediately

src/testcases/run_in_band/parameters.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      await daoMember1.submitUpdateParamsFeeburnerProposal(
+        'Proposal #3',
+        'Software upgrade proposal. Will pass',
+        dao.contracts.voting.address,
+        '1000',
+      );
+    });
+
+    describe('vote for proposal #3', () => {
+      const proposalId = 3;
```
sotnikov-s, line 156: how do you find an idea of introducing proposals counter like `lastProposalId` which will be incremented each time you make a proposal and reused as is instead of hardcoded digits? like that, it will be easily possible to rearrange tests and add more to the middle just with a simple copy+paste. it will also reduce the hardcode part. just an idea very much unnecessary to be implemented, totally up to you

sotnikov-s (changes requested):
besides the new comments, I still see a number of misprints/ambiguities/unaccounted comments from the previous time

src/testcases/run_in_band/parameters.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      });
+      test('execute passed proposal', async () => {
+        paramsBefore = await neutronChain.queryTokenfactoryParams();
+        await daoMember1.executeProposalWithAttempts(proposalId);
+      });
+      test('check if params changed after proposal execution', async () => {
+        const paramsAfter = await neutronChain.queryTokenfactoryParams();
+        // before: {"params":{"denom_creation_fee":[],
+        // "denom_creation_gas_consume":"0",
+        // "fee_collector_address":"neutron1suhgf5svhu4usrurvxzlgn54ksxmn8gljarjtxqnapv8kjnp4nrstdxvff"}}
```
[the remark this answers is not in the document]
quasisamurai: Default params are defined at `params.go`. This comment purpose is just to make test more clear when you read it

src/testcases/run_in_band/parameters.test.ts
quasisamurai: Purpose of this test is to make sure that it is possible to update params via governance, so i don't think it would make any sense (until we change the whole idea of this test suit) to rearrange order/insert smth in the middle. But idea w non-hardcoded `id` is pretty nice, i'll add it.

src/testcases/run_in_band/parameters.test.ts
quasisamurai: This is just a common way how we call this type of funcs (e.g. exaclty same way is `subdao/gov` test suits. I personally  don't see any issue w that

quasisamurai:
https://github.com/neutron-org/neutron-tests/actions/runs/6533928970

sotnikov-s (approved):

PR #278, opened 2024-04-04
fix: PCL lockdrop LP early staking

sotnikov-s:
#### task:
https://hadronlabs.atlassian.net/browse/NTRN-258

#### this PR
- adds an additional TGE participant that has only one lockup in atom pool. The user also participates in liquidity migration. Read the reasoning for the user in the next point
- expands the liquidity migration process to three steps:
      - migration of `airdropAuctionLockdropVestingMigration` (atom + usdc)
      - migration of `cmInstantiator` (atom + usdc, it's the last usdc pool participant, so the pool gets staked, but atom pool is not yet, and the user then withdraws atom position from not yet staked pool)
      - migration of `airdropAtomAuctionLockdropMigration` (atom, it's the last atom pool participant, so the pool gets staked)
- introduces several little fixes and additional tests

#### related PRs:
- https://github.com/neutron-org/neutron-tge-contracts/pull/89

### DIFF
--- a/src/testcases/parallel/overrule.test.ts
+++ b/src/testcases/parallel/overrule.test.ts
@@ -16,7 +16,7 @@ import {
 
 const config = require('../../config.json');
 
-describe('Neutron / Subdao', () => {
+describe('Neutron / Subdao Overrule', () => {
   let testState: TestStateLocalCosmosTestNet;
   let neutronChain: CosmosWrapper;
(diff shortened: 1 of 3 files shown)


sotnikov-s:
https://github.com/neutron-org/neutron-tests/actions/runs/8568032470

PR #314, opened 2024-06-20
feat: whitelist tf hooks

jcompagni10:
for https://github.com/neutron-org/neutron/pull/587 -- Allow chain manager dao to whitelist tokenfactory hooks

**Related PRs**
https://github.com/neutron-org/neutron-dao/pull/107
https://github.com/neutron-org/neutron-sdk/pull/150
https://github.com/neutron-org/neutronjsplus/pull/43

**Integration Test Run**
https://github.com/neutron-org/neutron-tests/actions/runs/9765613806


sotnikov-s (approved):

pr0n00gler (dismissed):

src/testcases/parallel/tokenfactory.test.ts
```diff
@@ -13,19 +14,70 @@ import {
...
+  neutronChain: CosmosWrapper,
+  subDao: Dao,
+  subdaoMember1: DaoMember,
+  codeID: number,
+  denomCreator: string,
+) {
+  const chainManagerAddress = (await neutronChain.getChainAdmins())[0];
+  const proposalId = await subdaoMember1.submitUpdateParamsTokenfactoryProposal(
+    chainManagerAddress,
+    'Proposal #2',
```
sotnikov-s, line 43: given the fact you call this func several times, it's incorrect (and doesn't actually have any point to do so) to set the `#2` number to the proposal in the title

src/testcases/parallel/tokenfactory.test.ts
```diff
@@ -13,19 +14,70 @@ import {
...
+    'Cron update params proposal. Will pass',
```
sotnikov-s, line 44: it's not cron params update proposal. please give your changes a thorough look in the context of the change when copy+pasting

sotnikov-s (changes requested):

jcompagni10:
https://github.com/neutron-org/neutron-tests/actions/runs/9670374932

sotnikov-s (commented):
excuse me, one more thing. please add a test that asserts that hooks fail if not whitelisted

src/testcases/parallel/tokenfactory.test.ts
```diff
@@ -346,59 +346,69 @@ describe('Neutron / Tokenfactory', () => {
...
-      }>(contractAddress, {
-        sudo_result_block_before: {},
+      test('set non-whitelisted hook fails  ', async () => {
+        const res = await msgSetBeforeSendHook(
+          neutronAccount,
+          ownerWallet.address.toString(),
+          newTokenDenom,
+          contractAddress,
+        );
+        expect(res.code).not.toEqual(0); // set hook fails
```
sotnikov-s, line 356: would be more accurate to check that the exact error is returned, not that just something went wrong

sotnikov-s (changes requested):

src/testcases/parallel/tokenfactory.test.ts
```diff
@@ -197,6 +342,39 @@ describe('Neutron / Tokenfactory', () => {
...
+        'create_denom',
+        'new_token_denom',
+      );
+      const res2 = await msgSetBeforeSendHook(
+        neutronAccount,
+        ownerWallet.address.toString(),
+        newTokenDenom,
+        contractAddress,
+      );
+      expect(res2.code).toEqual(14); // "beforeSendHook is not whitelisted"
```
sotnikov-s, line 361: it's good enough to check the code, but just FYI for the future: a more convenient IMO way to check errors is by the error message e.g. like here (or actually in many other places in our tests) https://github.com/neutron-org/neutron-integration-tests/blob/601f0614b1c5e57dc7c75f91b4e20701fc4527ca/src/testcases/run_in_band/tge.auction.test.ts#L1977-L1988

src/testcases/parallel/tokenfactory.test.ts
```diff
@@ -13,19 +14,69 @@ import {
...
+  );
+
+  let timelockedProp = await subdaoMember1.supportAndExecuteProposal(
+    proposalId,
+  );
+  await waitSeconds(10);
+  await subdaoMember1.executeTimelockedProposal(proposalId);
+  timelockedProp = await subDao.getTimelockedProposal(proposalId);
+  expect(timelockedProp.id).toEqual(proposalId);
+  expect(timelockedProp.status).toEqual('executed');
```
sotnikov-s, line 66: didn't this (speeded up?) approach work? I can see that you added it and then removed in the latest commits
image

sotnikov-s (dismissed):
comments above are only informative, so approving it

src/testcases/parallel/tokenfactory.test.ts
jcompagni10: could be mistaken but the way `msgSetBeforeSendHook` is called I can't use `rejects`

PR #318, opened 2024-06-27
feat: rewrite simple test #ntrn-343

NeverHappened:
https://hadronlabs.atlassian.net/browse/NTRN-343

- Rewritten simple.test in a way that does not use any wrappers in neutronjsplus (actually does not use it for the most part!)
- encapsulated contract's upload/execute/instantiate/migrate in a WasmWrapper (without unnecessary query methods)
- moved constants that belong to tests only to neutron-integration-tests
- renamed rpc1 -> rpcNeutron and so on


swelf19 (approved):

quasisamurai (approved):

sotnikov-s (commented):
what do you think about renaming the simple test to something like miscellaneous, structuring it in a readable way, and moving the IBC part from it to a separate test file?

src/helpers/wait.ts
```diff
@@ -0,0 +1,27 @@
+import { CosmWasmClient } from '@cosmjs/cosmwasm-stargate';
+import { StargateClient } from '@cosmjs/stargate';
+import { sleep } from '@neutron-org/neutronjsplus/dist/wait';
+
+export const waitBlocks = async (
```
sotnikov-s, line 5: can we merge the waitBlocks and getWithAttempts funcs into a single file? I find it too much to create a file per a tiny function. we could merge them into a `misc.ts` file here that would contain various tiny helpers like these ones. the other introduced files look good for me in this regard

src/testcases/parallel/simple.test.ts
```diff
@@ -100,85 +108,110 @@ describe('Neutron / Simple', () => {
...
-          gaiaAccount.wallet.address,
-          {
-            revisionNumber: BigInt(2),
-            revisionHeight: BigInt(100000000),
-          },
+        const fee = {
+          gas: '200000',
+          amount: [{ denom: NEUTRON_DENOM, amount: '1000' }],
+        };
+        const res = await neutronClient.client.signAndBroadcast(
```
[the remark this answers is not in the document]
NeverHappened: it's not clear for me how to do this. There is indeed message clients but they do not have link to any sender and signer, and docs does not have info on that. I'll look into that more

src/helpers/wasmClient.ts
```diff
@@ -0,0 +1,108 @@
...
+  const client = await SigningCosmWasmClient.connectWithSigner(
+    rpc,
+    wallet.directwallet,
+    { registry },
+  );
+  return new WasmWrapper(wallet, client, registry, CONTRACTS_PATH, denom);
+}
+
+// WasmWrapper simplifies cosmwasm operations for tests
+export class WasmWrapper {
```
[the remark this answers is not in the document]
NeverHappened: This is a good point. It's not that easy though. I'll try to do that.

src/testcases/parallel/simple.test.ts
```diff
@@ -1,89 +1,97 @@
...
-    );
-    neutronAccount = await createWalletWrapper(
-      neutronChain,
-      await testState.walletWithOffset('neutron'),
+      new Registry(neutronTypes),
     );
-    gaiaChain = new cosmosWrapper.CosmosWrapper(
+    gaiaAccount = await testState.nextWallet('cosmos');
+    gaiaAccount2 = await testState.nextWallet('cosmos');
+    gaiaClient = await wasm(
```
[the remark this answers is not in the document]
swelf19: btw, gaia integrated wasmd in the latest version (governance gated for now)

sotnikov-s (approved):

NeverHappened:
test run: https://github.com/neutron-org/neutron-tests/actions/runs/9874161701/job/27267981590

src/helpers/local_state.ts
```diff
@@ -5,99 +5,103 @@ import { CosmosWrapper } from '@neutron-org/neutronjsplus/dist/cosmos';
...
-    if (this.taken[prefix][idx]) {
-      return this.randomWallet(prefix);
-    }
-    this.taken[prefix][idx] = true;
-    return mnemonicToWallet(this.mnemonics[idx], prefix);
-  }
+  // Returns new wallet for a given `network`.
+  // The wallet is prefunded in a globalSetup.
+  // That way we can safely use these wallets in a parallel tests
+  // (no sequence overlapping problem when using same wallets in parallel since they're all unique).
```
swelf19, line 97: overlap still may happens if i call `nextWallet` more then `WALLETS_PER_TEST_FILE` times in a test. Right? The odds are low but if i right, may be it worth to mention it?
We can check if `currentOffset` > `WALLETS_PER_TEST_FILE` and give a warning

src/helpers/local_state.ts
```diff
@@ -5,99 +5,103 @@ import { CosmosWrapper } from '@neutron-org/neutronjsplus/dist/cosmos';
...
 
-    this.rest1 = restNeutron;
-    this.rest2 = restGaia;
+    this.currentIndexes = { neutron: {}, cosmos: {} };
+  }
 
-    this.icqWebHost = 'http://localhost:9999';
+  async init() {
+    if (this.suite) {
+      this.offset = await testFilePosition(this.suite);
```
swelf19, line 72: is the offset an offset of the test in a list? Not quite clear. Suggestion to rename it with a meaningful name.

swelf19 (approved):

src/helpers/local_state.ts
NeverHappened: I think we can throw in this case, as this is not good

swelf19 (approved):

PR #324, opened 2024-07-05
Feat: slinky tests to reflect market map creation in genesis

pr0n00gler:
We don't need **query market with empty metadata_JSON** as a separate testcase, since each metadata_json in marketmap is empty now

Related PR:
* https://github.com/neutron-org/neutron/pull/614


pr0n00gler:
https://github.com/neutron-org/neutron-tests/actions/runs/9818715409

NeverHappened (approved):

PR #347, opened 2024-09-10
Rate limit inttests

quasisamurai:
This is very basic tests for Ibc rate limit module to ensure it works as expected in our environment
[neutronjsplus PR](https://github.com/neutron-org/neutronjsplus/pull/64)


package.json
```diff
@@ -30,6 +30,7 @@
     "test:dex_bindings": "NO_WAIT_CHANNEL1=1 NO_WAIT_HTTP2=1 NO_WAIT_CHANNEL2=1 NO_WAIT_DELAY=1 vitest --run src/testcases/run_in_band/dex_bindings --bail 1",
     "test:chain_manager": "vitest --run src/testcases/run_in_band/chain_manager --bail 1",
     "test:feemarket": "NO_WAIT_CHANNEL1=1 NO_WAIT_HTTP2=1 NO_WAIT_CHANNEL2=1 NO_WAIT_DELAY=1 vitest --run src/testcases/run_in_band/feemarket --bail 1",
+    "test:rate_limit": "vitest --run src/testcases/run_in_band/rate_limit --bail 1",
```
sotnikov-s, line 33: let's explicitly specify that this is about IBC rate limiting? here and everywhere across the repo

setup/hermes/config.toml
```diff
@@ -102,7 +102,7 @@ store_prefix = 'ibc'
 default_gas = 100000
 max_gas = 3000000
 gas_price = { price = 0.0025, denom = 'untrn' }
-gas_multiplier = 2.0
+gas_multiplier = 2.7
```
sotnikov-s, line 105: why? because of computation increase during packet_recv and packet_send handlers?

src/helpers/signing_neutron_client.ts
```diff
@@ -228,3 +235,17 @@ export class SigningNeutronClient extends CosmWasmClient {
     );
   }
 }
+
+export async function CreateWalletFromMnemonic(
```
sotnikov-s, line 232: why don't use the already existing mnemonicToWallet from helpers/local_state.ts? pretty similar

src/testcases/parallel/ibc_transfer.test.ts
```diff
@@ -483,7 +483,7 @@ describe('Neutron / IBC transfer', () => {
         hermes checks height on remote chain and Timeout error occurs.
         */
         const currentHeight = await gaiaClient.getHeight();
-        await waitBlocks(15, gaiaClient);
+        await waitBlocks(25, gaiaClient);
```
sotnikov-s: are you sure that this and several other pause increases are necessary? another tests run duration increase

src/helpers/signing_neutron_client.ts
```diff
@@ -228,3 +228,17 @@ export class SigningNeutronClient extends CosmWasmClient {
...
+  mnemonic: string,
+  prefix: string,
+): Promise<DirectSecp256k1HdWallet> {
+  const wallet = await DirectSecp256k1HdWallet.fromMnemonic(mnemonic, {
+    prefix: prefix,
+  });
+
+  const accounts = await wallet.getAccounts();
+
+  console.log('Wallet address:', accounts[0].address);
```
sotnikov-s, line 242: don't forget to remove debug logs

src/testcases/parallel/ibc_transfer.test.ts
```diff
@@ -572,7 +578,14 @@ describe('Neutron / IBC transfer', () => {
...
 
-        await neutronClient.waitBlocks(5);
+        await neutronClient.waitBlocks(35);
+
+        // check that failures count is the same
+        failuresResAfter = await contractManagerQuerier.failures({
+          address: ibcContract,
+        });
+
+        console.log(failuresResAfter);
```
sotnikov-s: more debug logs and operations to remove

sotnikov-s (changes requested):

setup/hermes/config.toml
quasisamurai: yes, otherwise during `ibc_transfer` test contract isn't refunded because of out of gas

src/helpers/signing_neutron_client.ts
quasisamurai: haven't seen it, thx

src/testcases/parallel/ibc_transfer.test.ts
quasisamurai: It seems like yes (i found it last week), but i'll recheck this once again bc nobody likes long testruns

setup/hermes/config.toml
sotnikov-s: I wonder if it's got required to increase the multiplier because of a poor simulation process or because the IBC transactions reached the ceiling of `max_gas = 3000000`. if the latter, why don't increase the `max_gas` instead of the multiplier? would be cleaner

src/testcases/run_in_band/rate_limit.test.ts
```diff
@@ -0,0 +1,427 @@
...
+            },
+          ],
+          fee,
+        );
+        expect(res.code).toEqual(0);
+      });
+    });
+    describe('with limit, Gaia -> Neutron', () => {
+      test('check that weird IBC denom is uatom indeed', async () => {
+        const resBefroreLimit = await gaiaClient.signAndBroadcast(
```
sotnikov-s, line 276: why do you need an IBC transfer action in a `check that weird IBC denom is uatom indeed` test scope whereas the IBC transfer result is not even used in the check? would be cleaner to split these to two different test scopes e.g.
1. `check that weird IBC denom is uatom indeed`
2. `IBC transfer without any limits` (as above for the outgoing transfer)

src/testcases/run_in_band/rate_limit.test.ts
```diff
@@ -0,0 +1,427 @@
...
+          604800,
+          1,
+          1,
+        );
+        await neutronClient.execute(rlContract, {
+          add_path: quota,
+        });
+      });
+
+      // Note: we don't unset the limit afterwards, instead we're removing rate limiting contract from params
```
sotnikov-s, line 320: would be better to move (and maybe rephrase) the comment to the place you make the removal as an explanation to the removal action

src/testcases/run_in_band/rate_limit.test.ts
```diff
@@ -0,0 +1,427 @@
...
+      test('IBC transfer to limit', async () => {
```
sotnikov-s, line 321: the name of the test is not correct

src/testcases/run_in_band/rate_limit.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        const uatomibcSupply = await bankQuerier.supplyOf({
+          denom: UATOM_IBC_TO_NEUTRON_DENOM,
+        });
+
+        const amount = (
+          BigInt(uatomibcSupply.amount.amount) / BigInt(100) +
+          BigInt(1)
+        ).toString();
```
sotnikov-s, line 329: you make a beyond-the-limit transfer while not making a 1%-1 as above. could you please add that good limited case check?

src/testcases/run_in_band/rate_limit.test.ts
```diff
@@ -0,0 +1,427 @@
...
+      });
+      test('perform IBC send after removig of contract: should be fine', async () => {
+        const res = await neutronClient.signAndBroadcast(
+          [
+            {
+              typeUrl: NeutronMsgTransfer.typeUrl,
+              value: NeutronMsgTransfer.fromPartial({
+                sourcePort: 'transfer',
+                sourceChannel: TRANSFER_CHANNEL,
+                token: { denom: UATOM_IBC_TO_NEUTRON_DENOM, amount: '10' },
```
sotnikov-s, line 387: given the amount, this IBC send would work even without the rate limiter contract removal since it doesn't go beyond the limit

sotnikov-s (changes requested):
would be great to see the following additional test scenarios:
- quota reset, i.e. IBC transfer limits are reset and transfers work again after a quota period is renewed
- multiple quotas of different limits and time spans, i.e. if you have a daily and a weekly quotas of 5 and 10 respectfully, even if a daily quota is reached twice and then renewed, the weekly quota should still prevent transfers
- if the quota is reached by a sender, other senders also cannot make more sends
- IBC transfers made by smart contracts via the ibc-transfer module are also limited

pr0n00gler:
> * IBC transfers made by smart contracts via the ibc-transfer module are also limited

This is a very good point.

But i don't think we need to implement others if they are covered in unit-tests in Neutron

src/testcases/run_in_band/rate_limit.test.ts
quasisamurai: this test is testing a different thing: is that after removal of contract from the state ibc send simply works. it has nothing to do with a limits themselves (because rate-limiting MW action is completely removed from the ibc stack)

src/testcases/run_in_band/rate_limit.test.ts
quasisamurai: I pesonally don't see any reason to reproduce _an exact_ same test scenario, but for different denom. the logic of that is no different from first test case, so here in order to persuade slight diversity of scenarios we're hitting limit in one tx

src/testcases/run_in_band/rate_limit.test.ts
quasisamurai: split in 2 for better readability

src/testcases/run_in_band/rate_limit.test.ts
sotnikov-s: ah, I thought it was an incoming transfer test scenario. alright then, this is fine. but what about an incoming transfer test scenario? I think it makes sense to add one

src/testcases/run_in_band/rate_limit.test.ts
sotnikov-s: why don't cover two things in a single operation then: not that just IBC transfer works, but that the rate limiting is not applied anymore? by simply removing the contract from the state and making a tiny send you won't be able to know if it's even been removed and is not involved in the IBC middleware stack

src/testcases/run_in_band/rate_limit.test.ts
sotnikov-s: what does `to limit` mean then if you're overcoming the limit?

src/testcases/run_in_band/ibc_rate_limit.test.ts
```diff
@@ -0,0 +1,515 @@
...
+import { createRPCQueryClient as createNeutronClient } from '@neutron-org/neutronjs/neutron/rpc.query';
+import { NeutronQuerier } from '@neutron-org/neutronjs/querier_types';
+import { QueryClientImpl as IbcQueryClient } from '@neutron-org/neutronjs/ibc/applications/transfer/v1/query.rpc.Query';
+import { Wallet } from '../../helpers/wallet';
+
+const TRANSFER_CHANNEL = 'channel-0';
+const UATOM_IBC_TO_NEUTRON_DENOM =
+  'ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2';
+
+// These are th
```
sotnikov-s, line 32: These are th?

sotnikov-s (dismissed):
nice!

pr0n00gler (approved):

PR #350, opened 2024-09-13
fix: add ICQ test for new update period behaviour #NTRN-396

joldie777:
Task: https://hadronlabs.atlassian.net/browse/NTRN-396

Related PRs:
https://github.com/neutron-org/neutron-query-relayer/pull/93


joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/10848288468

src/testcases/run_in_band/interchain_kv_query.test.ts
```diff
@@ -69,6 +69,7 @@ describe('Neutron / Interchain KV Query', () => {
     4: 3,
     5: 4,
     6: 11,
+    7: 50,
```
pr0n00gler, line 72: Maybe set it to some veeeeery big number? E.g. 10_000_000 or smth.
Cause it think it's quite possible that because of some lag in tests we can reach an update height with such small update period (50)

joldie777: Changed for 1_000_000

joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/10901691854

pr0n00gler (dismissed):

src/testcases/run_in_band/interchain_kv_query.test.ts
```diff
@@ -351,6 +351,19 @@ describe('Neutron / Interchain KV Query', () => {
           testState.wallets.cosmos.val1.address,
         );
       });
+
+      test('register icq #7: balance', async () => {
+        const updatePeriod = (await neutronClient.getHeight()) + 1_000_000;
```
pr0n00gler, line 356: Why is there `updatePeriod` equals to `current_height + 1_000_000`?
Why not just million?

joldie777: This is needed to make sure that the ICQ relayer behaves as expected. We need to check that query result will be submitted before the first update period passes.
I mean if the current block number would be greater than the update period (e.g. 1_000_005), the relayer would run the query immediately and [this](https://github.com/neutron-org/neutron-integration-tests/blob/421240bd910bd650ab8acae213a52333a015d7e1/src/testcases/run_in_band/interchain_kv_query.test.ts#L489-L491) check would fail.

pr0n00gler: But to reach 1M height the test needs to run for a couple of days, no?

pr0n00gler: So nothing will be broken if update_height equals just to 1M, right?

joldie777: You're right! I'll fix

joldie777:
Test run: https://github.com/neutron-org/neutron-tests/actions/runs/10941015644

pr0n00gler (approved):

PR #352, opened 2024-09-17
Feat: software upgrade via chain manager [NTRN-400]

sotnikov-s:
### task
https://hadronlabs.atlassian.net/browse/NTRN-400

### this PR
- adds tests for both `MsgSoftwareUpgrade` and `MsgCancelSoftwareUpgrade` adminmodule messages being sent via respestive introduced chain manager strategies;
- refactors the cron `MsgAddSchedule` and `MsgRemoveSchedule` tests so they don't have SubDAO involved but an ordinary account. Decided to do that when handled accidentally failing `add schedule` test with a totally uninformative and untraceable error `expected executed, got execution_failed` on `AddSchedule` SubDAO proposal execution to 1) get rid of poor error; 2) speed the tests up (interaction with chain manager from an ordinary account is much faster than from the SubDAO). Viewing and understanding the cron-related diff will be much easier if you first check out the way the upgrade-related tests are written because the new cron-related tests are written the same way.

### related PRs:
- https://github.com/neutron-org/neutron-dao/pull/115
- https://github.com/neutron-org/neutron/pull/727

### DIFF
--- a/src/testcases/run_in_band/chain_manager.test.ts
+++ b/src/testcases/run_in_band/chain_manager.test.ts
@@ -416,7 +416,6 @@ describe('Neutron / Chain Manager', () => {
 
     const scheduleName = 'schedule1';
     test('add cron schedule', async () => {
-      await neutronClient.waitBlocks(5);
       await cronAccessClient.execute(
         chainManagerAddress,
         {
--- a/src/testcases/run_in_band/chain_manager.test.ts
+++ b/src/testcases/run_in_band/chain_manager.test.ts
@@ -19,9 +19,12 @@ import { setupSubDaoTimelockSet } from '../../helpers/dao';
 import { QueryClientImpl as CronQueryClient } from '@neutron-org/neutronjs/neutron/cron/query.rpc.Query';
 import { QueryClientImpl as AdminQueryClient } from '@neutron-org/neutronjs/cosmos/adminmodule/adminmodule/query.rpc.Query';
 import { QueryClientImpl as TokenfactoryQueryClient } from '@neutron-org/neutronjs/osmosis/tokenfactory/v1beta1/query.rpc.Query';
+import { QueryClientImpl as UpgradeQueryClient } from '@neutron-org/neutronjs/cosmos/upgrade/v1beta1/query.rpc.Query';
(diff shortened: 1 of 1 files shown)


sotnikov-s:
https://github.com/neutron-org/neutron-tests/actions/runs/10911240382

src/testcases/run_in_band/chain_manager.test.ts
```diff
@@ -447,4 +467,363 @@ describe('Neutron / Chain Manager', () => {
...
+          expect(plan.plan).toBeUndefined();
+        });
+      });
+    });
+  });
+});
+
+// returns a wasm execute message to a chain manager that adds cron schedules permissions to
+// a given authority.
+const buildAddCronSchedulesStrategyMsg = (
```
pr0n00gler, line 714: Maybe this (and all the helpers above) should be located in `helpers.ts` file or smth and available for the whole testsuit?

pr0n00gler (approved):
LGTM, just a small question

src/testcases/run_in_band/chain_manager.test.ts
sotnikov-s: do you think there is a chance they might be reused in other test files?

pr0n00gler: Sure, they look like something that can be used a lot.

E.g. Nicholas is (was dunno) working on a contract that is being called by cron module, so he needs(ed) a lot of cron proposals for his test scenarios

sotnikov-s: in the linked PR the schedules are added/removed via DAO proposals, not directly via chain manager, so they can't just be reused there. re Nicolas: idk what he's doing or was doing, so I think we better let him know about such helpers and he could publish and reuse them if needed.

also I think, if publishing chain manager helpers, we better create a chain manager client that would expose all possible proposal creation helpers e.g. params changes, software updates, cron schedules. these helpers could then be reused in DAO calls (like in the linked PR), and the helpers like `daoMember.submitAddSchedule` are IMO too specific and can be removed from neutronjsplus and replaced with simple `daoMember.submitSingleChoiceProposal(chainManagerClient.buildAddCronSchedulesStrategyMsg())`

PR #387, opened 2025-06-05
test removing validator

NeverHappened:


src/helpers/staking.ts
```diff
@@ -165,6 +176,7 @@ export const undelegateTokens = async (
     ],
     { amount: [{ denom: NEUTRON_DENOM, amount: '5000000' }], gas: '2000000' },
   );
+  console.log('undelegate res: ' + JSON.stringify(res.rawLog));
```
pr0n00gler, line 179: I don't think we need this log in production

src/helpers/staking.ts
```diff
@@ -101,6 +101,17 @@ export const getBondedTokens = async (
   return bondedTokens;
 };
 
+export const getTrackingValidators = async (
+  client: SigningNeutronClient,
+  stakingTrackerAddr: string,
+): Promise<any> => {
+  const validators = await client.queryContractSmart(stakingTrackerAddr, {
+    list_validators: { limit: 1000 },
```
pr0n00gler, line 109: Let's make limit optional argument with a default value for the method?
It can be useful in the future, since you are introducing the method

pr0n00gler (changes requested):

src/helpers/staking.ts
```diff
@@ -101,6 +101,17 @@ export const getBondedTokens = async (
   return bondedTokens;
 };
 
+export const getTrackingValidators = async (
```
sotnikov-s, line 104: I think it should be getTrack**ed**Validators for 1) grammar reasons; 2) we have `getTrackedStakeInfo` func above

src/testcases/run_in_band/staking-tracker.test.ts
```diff
@@ -112,6 +117,95 @@ describe('Neutron / Staking Tracker - Extended Scenarios', () => {
...
+      const moduleValidatorsBefore = await stakingQuerier.validators({
+        status: '',
+      });
+      expect(trackingValidatorsBefore.length).toEqual(
+        moduleValidatorsBefore.validators.length,
+      );
+      expect(trackingValidatorsBefore.length).toEqual(3);
+
+      // undelegate all tokens
+      const valAddr = validatorsRes.validators[0].operatorAddress;
```
sotnikov-s, line 191: why don't use `wallet.valAddress` here? would be much more predictable

src/testcases/run_in_band/staking-tracker.test.ts
```diff
@@ -112,6 +117,95 @@ describe('Neutron / Staking Tracker - Extended Scenarios', () => {
...
+        client,
+        STAKING_TRACKER,
+      );
+      const moduleValidatorsBefore = await stakingQuerier.validators({
+        status: '',
+      });
+      expect(trackingValidatorsBefore.length).toEqual(
+        moduleValidatorsBefore.validators.length,
+      );
+      expect(trackingValidatorsBefore.length).toEqual(3);
```
sotnikov-s, line 188: any reasons to check equality to 3? if we change the initial number of validators in the future, this test will have to be fixed. why don't just check that `trackingValidatorsBefore.length == moduleValidatorsBefore.validators.length` here and check that the numbers decreased by 1 after undelegation?

sotnikov-s (changes requested):

src/testcases/run_in_band/staking-tracker.test.ts
NeverHappened: agreed

pr0n00gler (approved):

